PR #946, opened 2013-06-14
Add move tab to tab context menu

char101:
Add a "Move To" submenu to tab context menu that enables the user to move the tab to another window. The menu contains the list of opened windows excluding the current window.

movetab

If the last tab is moved, then the current window is closed and the other window is raised.

I actually want to made this a separate pull request but forget to switch branch before pushing:
Middle clicing on the tab list menu closes the tab. The menu item is grayed out instead of deleted so that the menu does not close and also to give feedback the the user on what tabs have been closed.

closetab


nowrep:
This functionality is now implemented with drag&drop b7341474d60fd65df0e2dfcb8fb9a40f0f48ae67

PR #1402, opened 2014-06-22
Hide empty bookmarks folder in menu

kulti:
As for me, I think, that empty folders not needed in bookmarks menu. But I don't know how other people looks on it. So I added setting to save current behavior.


minlexx:
I also find that empty "Unsorted bookmarks" folder annoying...

PR #1577, opened 2015-01-02
Add 'Block domain' button to the cookie manager

gzsombor:
So it is much simpler to remove a cookie, and get rid of a particularly offending cookie dropper site

### DIFF
--- a/src/lib/cookies/cookiemanager.cpp
+++ b/src/lib/cookies/cookiemanager.cpp
@@ -56,6 +56,7 @@ CookieManager::CookieManager(QWidget* parent)
     connect(ui->cookieTree, SIGNAL(currentItemChanged(QTreeWidgetItem*,QTreeWidgetItem*)), this, SLOT(currentItemChanged(QTreeWidgetItem*,QTreeWidgetItem*)));
     connect(ui->removeAll, SIGNAL(clicked()), this, SLOT(removeAll()));
     connect(ui->removeOne, SIGNAL(clicked()), this, SLOT(removeCookie()));
+    connect(ui->blockDomain, SIGNAL(clicked()), this, SLOT(blockCurrentHostAndRemoveCookie()));
     connect(ui->close, SIGNAL(clicked(QAbstractButton*)), this, SLOT(close()));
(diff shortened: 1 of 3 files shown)


gzsombor:
What do you think?

nowrep:
The feature is ok, but I don't really like the code.
Please see the in-line notes.

src/lib/cookies/cookiemanager.h
```diff
@@ -62,6 +63,7 @@ private slots:
 private:
     void closeEvent(QCloseEvent* e);
     void keyPressEvent(QKeyEvent* e);
+    void addBlacklist(QString server);
```
nowrep: `const QString &server`

src/lib/cookies/cookiemanager.cpp
```diff
@@ -108,17 +109,19 @@ void CookieManager::removeAll()
     ui->cookieTree->clear();
 }
 
-void CookieManager::removeCookie()
+// Return the domain of the removed cookie
+QString CookieManager::removeCookie()
```
nowrep: Instead of changing the `removeCookie()` method, there should be new method to return domain of current cookie.
Or, as it is only used in one place, it may not even be made to function.

src/lib/cookies/cookiemanager.cpp
```diff
@@ -137,9 +141,17 @@ void CookieManager::removeCookie()
...
             ui->cookieTree->deleteItem(parentItem);
         }
+        domain = cookie.domain();
     }
 
     mApp->cookieJar()->setAllCookies(allCookies);
+    return domain;
+}
+
+void CookieManager::blockDomain()
```
nowrep: This looks confusing. Blocking domain should not mean to remove current cookie + add domain to blacklist. It should just add cookie to blacklist.

gzsombor:
Thanks for the comments, I've tried to address the issues with the new patch, but I'm not convinced, that this separate codepath to get the domain name in 'blockCurrentHostAndRemoveCookie' is a better approach.

nowrep:
This looks better, thanks :)

gzsombor:
Your welcome!

PR #1630, opened 2015-03-28
Use https for updating CA Certificates, prevent useless build of updater

Nikoli:
Both commits are tested only in Linux, so it means i did not check if updating actually works.

P.S.  About ```Will probably work only for Debian based distros``` comment from ```scripts/make_ca_bundle.sh```

```/etc/ssl/certs/ca-certificates.crt``` exists in Gentoo too:
```
$ du -sh /etc/ssl/certs/ca-certificates.crt src/lib/data/data/ca-bundle.crt
270K    /etc/ssl/certs/ca-certificates.crt
264K    src/lib/data/data/ca-bundle.crt
```

It is created by ```/usr/sbin/update-ca-certificates``` tool from ```app-misc/ca-certificates``` ebuild:
https://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/app-misc/ca-certificates/ca-certificates-20141019.3.17.4.ebuild?view=markup&revision=1.1

### DIFF
--- a/src/lib/app/qzcommon.cpp
+++ b/src/lib/app/qzcommon.cpp
@@ -30,4 +30,5 @@ QUPZILLA_EXPORT const char* AUTHOR = "David Rosca";
 QUPZILLA_EXPORT const char* COPYRIGHT = "2010-2014";
 QUPZILLA_EXPORT const char* WWWADDRESS = "http://www.qupzilla.com";
 QUPZILLA_EXPORT const char* WIKIADDRESS = "https://github.com/QupZilla/qupzilla/wiki";
+QUPZILLA_EXPORT const char* RAWVCS = "https://raw.githubusercontent.com/QupZilla/qupzilla/v1.8";
 }
--- a/src/lib/app/qzcommon.h
+++ b/src/lib/app/qzcommon.h
@@ -69,6 +69,7 @@ QUPZILLA_EXPORT extern const char* AUTHOR;
 QUPZILLA_EXPORT extern const char* COPYRIGHT;
 QUPZILLA_EXPORT extern const char* WWWADDRESS;
 QUPZILLA_EXPORT extern const char* WIKIADDRESS;
+QUPZILLA_EXPORT extern const char* RAWVCS;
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by nowrep** [comment]
Sorry for the late reply.

I'm not really sure on this one, the github download url seems to change from time to time. I do however like the idea to download it from https (which is not available on qupzilla.com).

**Comment by Nikoli** [comment]
The urls are still working, so may be github does not change them too often? :)

**Comment by nowrep** [comment]
Yep, it didn't change it now for some time :D

**Comment by Nikoli** [comment]
Also what about ```Do not build CA bundle updater when not used ``` commit, is it fine?

**Comment by nowrep** [comment]
Yeah, that one +1

File: src/lib/app/qzcommon.cpp [hunk 1]
```diff
@@ -30,4 +30,5 @@ QUPZILLA_EXPORT const char* AUTHOR = "David Rosca";
 QUPZILLA_EXPORT const char* COPYRIGHT = "2010-2014";
 QUPZILLA_EXPORT const char* WWWADDRESS = "http://www.qupzilla.com";
 QUPZILLA_EXPORT const char* WIKIADDRESS = "https://github.com/QupZilla/qupzilla/wiki";
+QUPZILLA_EXPORT const char* RAWVCS = "https://raw.githubusercontent.com/QupZilla/qupzilla/v1.8";
```
nowrep: Remove the url here and move it to cabundleupdater.cpp - preferably as 
```cpp
static const char* CA_BUNDLE_URL = "https://raw.githubusercontent.com/QupZilla/qupzilla/v1.8/src/lib/data/data/bundle_version";
```
and then it's good to go.

Nikoli:
Are you sure? I intentionally kept base url in src/lib/app/qzcommon.cpp because it can be used somewhere else too: for updating list of adblock subscriptions, for example.

nowrep:
Yes, I am sure. The url can currently only be used to download the certificates.

nowrep:
qupzilla.com now uses https, so this change is no longer needed.

PR #1655, opened 2015-05-14
fix adblock icon animation

vsegdaskray:
When adblock block popup window icon begin animate, by replacing on `QPixmap()` and return adblock.png.  When icon hide total width of elements decrease on 16px and record with domain name and ip address move to right on 16 px, after 500ms icon show and record move to left to 16 px.
i think this is unexpected behavior. So i create clear png( 16x16 ) and icon replaced by this image, not `QPixmap()`. Total width of elements not change and we see how icon blinking correctly


nowrep:
Instead of this, just set fixed size of the adblock icon.

vsegdaskray:
Sorry, i'm newbie in Qt and don't know that can do this and how do it correctly

vsegdaskray:
`setMinimumSize( );` when icon create fix it in 1 string, best solution as i think

PR #1691, opened 2015-08-10
Update BUILDING

kloun:


nowrep:
Any reason for this change?

kloun:
in openbsd, /usr/local is a extra soft PATH.

2015-08-10 20:45 GMT+03:00 David Rosca <notifications@github.com>:

PR #1701, opened 2015-09-10
Fix not closing cmdline files from /proc/ on read errors

philippwiesemann:
The cmdline files opened from /proc/ were not closed if reading the data from them failed (e.g. because the file was empty). This leaked the file descriptor which could lead to failures trying to open more files later.

This fault also exists on the master branch (but the functionality is currently not used there).

Thanks!


nowrep:
Thanks, can you make pull request also for `master` branch, or should I cherry-pick it?

PR #1702, opened 2015-09-11
Fix not closing cmdline files from /proc/ on read errors (in master branch)

philippwiesemann:
The cmdline files opened from /proc/ were not closed if reading the data from them failed (e.g. because the file was empty). This leaked the file descriptor which could lead to failures trying to open more files later.

See #1701.

Thanks!


nowrep:
Great, thanks!

PR #1705, opened 2015-09-13
Make AppData translatable

### LINKED ISSUE #1704: [i18n] Make AppData file translatable
~~Consider writing and installing an AppData file with the application description and some screenshots, else QupZilla would look bad in the GNOME and KDE Software Centers.~~

~~See these links for details.~~
~~http://people.freedesktop.org/~hughsient/appdata/~~
~~https://github.com/hughsie/appdata-tools~~

GreenLunar:
Partially fixes #1704

### DIFF
--- a/linux/appdata/qupzilla.appdata.xml
+++ b/linux/appdata/qupzilla.appdata.xml
@@ -6,15 +6,15 @@
  <name>QupZilla</name>
  <summary>Web Browser</summary>
  <description>
-  <p>
+  <_p>
     QupZilla is a new and very fast QtWebKit browser. It aims to be a lightweight web browser available through all major platforms. This project has been originally started only for educational purposes. But from its start, QupZilla has grown into a feature-rich browser.
-  </p>
-  <p>
+  </_p>
+  <_p>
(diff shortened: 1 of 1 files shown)


nowrep:
Why just partially? It makes it translatable, what more we need?

GreenLunar:
I do not know.  @hughsie may you please guide us?

GreenLunar:
@gottcode, how do you manage AppData files of your projects?

gottcode:
@GreenLunar I use a hackish solution. I wrote a Bash script to call ```intltool-extract``` to fetch the strings of the desktop and appdata files, then run ```xgettext``` on those files to create a .pot file, and then run ```msgmerge``` to update the .po files with new strings, and finally run ```intltool-merge``` to copy the translations back from the .po files into the desktop and appdata files. You can see the script here: https://github.com/gottcode/focuswriter/blob/master/resources/unix/po/update-po.sh

GreenLunar:
Thank you, Graeme.  I think such solution is acceptable, as [qupzilla.desktop](https://github.com/QupZilla/qupzilla/blob/v1.8/linux/applications/qupzilla.desktop) is also static.

GreenLunar:
Which means, if your solution will be accepted, that this commit is of no use.

gottcode:
I'm glad I could be of assistance.

PR #1709, opened 2015-09-16
Fix breaking RSS feed database by editing URLs

philippwiesemann:
If the URL of an already added feed was edited to contain certain errors (e.g. a single "%") the feed could not be changed or deleted afterwards.
The fault was that edited URLs were stored in the database without being converted to a QUrl first. This conversion silently corrects some errors in the URL which also changes its value. The URLs read from the database are converted to QUrl before being set on the tabs and later are used to find entries to update or delete. This failed for edited URLs because of the now different values.
This was fixed by converting URLs from QString to QUrl before storing them in the database. Being already converted the value is then not changed if being converted again after reading and can be used for finding the entries to update or delete in the database.


nowrep:
Thanks!

PR #1711, opened 2015-09-23
Fix not escaping characters for HTML directory listings (v1.8)

philippwiesemann:
The HTML pages created for FTP and file directory listings included the paths without escaping the HTML meta characters. The HTML generated for a path containing such characters was therefore not valid. This created problems when displaying or processing the page.

This should have been a rare problem because most paths do not contain such characters and the final display of the page (also the output from "Page Source") had most problems implicitly corrected (but not "<" followed by a letter).

Thanks!


nowrep:
Thanks for the patch.

PR #1712, opened 2015-09-23
Fix not escaping characters for HTML directory listings

philippwiesemann:
The HTML pages created for FTP and file directory listings included the paths without escaping the HTML meta characters. The HTML generated for a path containing such characters was therefore not valid. This maybe created problems when displaying or processing the page.

Thanks!


nowrep:
Those are not currently used with QtWebEngine, because Chromium has its own `file:` and `ftp:` handlers.

Issue #1719, opened 2015-10-15
Disabling/enabling plugins and menu item issue.


srazi:
When implementing what @nowrep said in this [comment](https://github.com/QupZilla/qupzilla-plugins/issues/42#issuecomment-108590421) I found an issue, please test following steps and report what happens here.

1- enable Example plugins,
2- you should see its menu item in `view=>sidebar` (please point there and see if you can see submenu)
3- disable Example plugins
4- again point to that submenu `view=>sidebar`

For me QupZilla crashes in step 4 (Qt4 and Qt 5.4.2)

nowrep:
Works fine in master.

nowrep:
It shouldn't crash, it uses QPointers and always check for null before using.

nowrep:
Is the `QPluginLoader::unload()` call successful?

```diff
diff --git a/src/lib/plugins/plugins.cpp b/src/lib/plugins/plugins.cpp
index 7aaaa63..70c3390 100644
--- a/src/lib/plugins/plugins.cpp
+++ b/src/lib/plugins/plugins.cpp
@@ -69,7 +69,7 @@ void Plugins::unloadPlugin(Plugins::Plugin* plugin)
     }
 
     plugin->instance->unload();
-    plugin->pluginLoader->unload();
+    qDebug() << plugin->pluginLoader->unload();
     emit pluginUnloaded(plugin->instance);
 
     m_availablePlugins.removeOne(*plugin);
```

srazi:
The issue occurs in `QObject:disconnect()` when calling `sender->metaObject()` that `sender` is sidebar action and metaObject is virtual.

**Console output:**
```python
TestPlugin::init called 
TestPlugin::unload called 
72 Plugins::unloadPlugin true 
(888.aac): Access violation - code c0000005 (first chance)
First chance exceptions are reported before any exception handling.
This exception may be expected and handled.
eax=0027c388 ebx=00000000 ecx=064ec580 edx=5de0b118 esi=13000007 edi=6690560a
eip=66b0f8e0 esp=0027c310 ebp=0027c3a4 iopl=0         nv up ei pl zr na pe nc
cs=001b  ss=0023  ds=0023  es=0023  fs=003b  gs=0000             efl=00010246
QtCored4!QObject::disconnect+0x290:
66b0f8e0 8b02            mov     eax,dword ptr [edx]  ds:0023:5de0b118=????????
```

**Backtrace:**
```python
[...]

PR #1720, opened 2015-10-15
Add plugins menu item to Tools menu.

srazi:
@nowrep would you please test if these commits work for you, unfortunately for me QZ crashes just like  #1719 .

### DIFF
--- a/src/lib/app/mainmenu.cpp
+++ b/src/lib/app/mainmenu.cpp
@@ -32,6 +32,7 @@
 #include "browsinglibrary.h"
 #include "clearprivatedata.h"
 #include "qzsettings.h"
+#include "pluginproxy.h"
 
 #include <QApplication>
 #include <QMetaObject>
@@ -447,6 +448,11 @@ void MainMenu::aboutToHideEditMenu()
 void MainMenu::aboutToShowToolsMenu()
 {
     m_actions[QSL("Tools/SiteInfo")]->setEnabled(true);
+
+    m_submenuExtensions->clear();
+    mApp->plugins()->populateExtensionsMenu(m_submenuExtensions);
+
+    m_submenuExtensions->menuAction()->setVisible(!m_submenuExtensions->actions().isEmpty());
 }
 
 void MainMenu::aboutToHideToolsMenu()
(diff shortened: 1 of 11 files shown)


nowrep:
And what about fixing that crash? :P

nowrep:
But indeed, it may crash only on Windows, because dynamic loading of libraries works slightly different than on Linux.

srazi:
@nowrep: did you mean it works for you?

> And what about fixing that crash? :P

Well it is strange. it crashes when calling `m_submenuExtensions->clear()` I also remove parent of QActions. again deleting them leads to crash.

nowrep:
I didn't test it yet.

src/lib/app/mainmenu.cpp
```diff
@@ -621,6 +627,10 @@ void MainMenu::init()
     ADD_ACTION("Tools/RssReader", m_menuTools, QIcon(), tr("RSS &Reader"), SLOT(showRssManager()), "");
     ADD_ACTION("Tools/WebInspector", m_menuTools, QIcon(), tr("Web In&spector"), SLOT(toggleWebInspector()), "Ctrl+Shift+I");
     ADD_ACTION("Tools/ClearRecentHistory", m_menuTools, QIcon::fromTheme(QSL("edit-clear")), tr("Clear Recent &History"), SLOT(showClearRecentHistoryDialog()), "Ctrl+Shift+Del");
+
+    m_submenuExtensions = new QMenu(tr("&Extensions"));
```
nowrep: Is there a leak? You always create the menu in aboutToHide, but never delete.

srazi: No.
`m_submenuExtensions` is created in `MainMenu::init()` and its actions are created as its child therefore `m_submenuExtensions->clear()` deletes them.

nowrep: Ah right, this is init(), i got confused with one line above (aboutToHideToolsMenu) :D

nowrep:
Yes, works fine here.

nowrep:
Also cherry-picked to `master`.

PR #1724, opened 2015-10-18
TabManager: Add an option to use it as main tabbar replacement.

srazi:
- Note: in `TabBar::setVisible()` we didn't need to
  check `(visible && m_window->isFullScreen())` because tabbar's
  visibility in fullscreen mode is managed by m_navigationContainer.


nowrep:
Well, that check was here to prevent showing tabbar when adding new tab in fullscreen, etc.

srazi:
Hum, adding a new tab in fullscreen will set tabar visible but relative to its parentWidget i.e. m_navigationContainer that is invisible.

Issue #1753, opened 2015-11-08
Last removed speed dial returns after reload


philippwiesemann:
This happens in QupZilla 1.8.8 (and also at least in 1.8.7).

Steps to reproduce the problem:
1. Start QupZilla
2. Go to "qupzilla:speeddial"
3. Remove all speed dials from page
4. Close now empty "qupzilla:speeddial" page
5. Go to "qupzilla:speeddial" again
6. Page is not empty and contains last removed speed dial again

It looks like the state of the page is not saved after the last speed dial was removed. The empty page seems to be serialized into an empty string and is therefore ignored (and maybe because else it would lead to the default state).

Thanks!

cranes-bill:
Sure, this is **Speed Dial**
What will be dialling, when everithing is missing!?
Also opening „Empty page“ can be set.

nowrep:
On the other hand, is this really an issue? There is not a use-case for empty speed dial page.

ke7ofi:
Actually, there could be. Since there's a built-in tool for showing custom colors/images on the Speed Dial, people might use it just for that reason.

philippwiesemann:
The remove button could be hidden on the last speed dial as a workaround (which would be fragile and not fixing the actual fault).
```css
div.entry:only-child .close { display: none; }
```

ke7ofi:
@philippwiesemann ah, the ol' bug-feature switcharoo. I think that's probably the best temporary solution as long as it stays temporary.

nowrep:
Actually, that workaround style looks fine. Can you please open a pull request?

ke7ofi:
While it restores consistency between what the user tells (and can tell) the browser to do and what it actually does, I feel like it's a bad thing to have a situation where the browser can't be returned entirely to its “blank” state without reinstallation.

PR #1760, opened 2015-11-12
SpeedDial: Hide remove button on last speed dial

philippwiesemann:
Hiding the remove button on the last speed dial prevents an empty page.

See Issue #1753.


nowrep:
Thanks

PR #1786, opened 2015-11-27
Fix duplicated RSS feed dialog having wrong parent

philippwiesemann:
The warning dialog about duplicated RSS feeds was opened with the main application window being the parent instead of the RSS feed window. This opened the warning dialog behind the RSS feed window which was confusing because the window needed to be moved away to find and close the dialog.

Thanks!

### DIFF
--- a/src/lib/rss/rssmanager.cpp
+++ b/src/lib/rss/rssmanager.cpp
@@ -165,7 +165,7 @@ void RSSManager::addFeed()
         return;
     }
 
-    addRssFeed(url, tr("New feed"), IconProvider::iconForUrl(url));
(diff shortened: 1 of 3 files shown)


nowrep:
What happens when this gets called from web notification? RSS Manager will be hidden, so `this` parent will be invalid.

Maybe add a `QWidget *parent` parameter to `addRss`?

philippwiesemann:
The commit was updated. Thanks for the suggestion.

This way is more safe than relying on the widget button being deactivated if the feed was already added.

nowrep:
Looks good now, thanks.

Issue #1735, opened 2015-10-25
Option to show only text in bookmarks bar


AcarBurak:
Please see the attached screenshots, Qupzilla's and Otter's bookmarks bars.
![qup](https://cloud.githubusercontent.com/assets/1426842/10718678/28ffc894-7b81-11e5-8a39-49fc458c13ab.png)
![otter](https://cloud.githubusercontent.com/assets/1426842/10718680/2ce975c2-7b81-11e5-9bdd-d295f2bd85b8.png)

PR #1802, opened 2015-12-10
Bookmarks: Add option to show only text in bookmarks bar

innermous:
Implements a feature, requested at #1735 .


nowrep:
It seems fine from brief look, but can you please make it against either `v1.8` or `master`? `qt5.5` is not active branch.

innermous:
Shoul

innermous:
Should i just create another pull request?

PR #1804, opened 2015-12-12
Fix adding Adblock list

### LINKED ISSUE #1801: Can't add Adblock list
My guess that is because of the **apostrophe** in the file name.
Steps to reproduce:
Go to [this](http://www.kiboke-studio.hr/i-dont-care-about-cookies/) page
Click on **Add to Adblock Plus** link
Click :ok: on confirmation dialogue
The result is the following:
capture du 2015-12-10 21 20 30

No problem with this with _V1.8 Qt4_

innermous:
Closes #1801 .

### DIFF
--- a/src/lib/adblock/adblockmanager.cpp
+++ b/src/lib/adblock/adblockmanager.cpp
@@ -147,7 +147,7 @@ void AdBlockManager::removeDisabledRule(const QString &filter)
 
 bool AdBlockManager::addSubscriptionFromUrl(const QUrl &url)
 {
-    const QList<QPair<QString, QString> > queryItems = QUrlQuery(url).queryItems();
+    const QList<QPair<QString, QString> > queryItems = QUrlQuery(url).queryItems(QUrl::FullyDecoded);
 
     QString subscriptionTitle;
     QString subscriptionUrl;


src/lib/adblock/adblockmanager.cpp
```diff
@@ -154,10 +154,14 @@ bool AdBlockManager::addSubscriptionFromUrl(const QUrl &url)
 
     for (int i = 0; i < queryItems.count(); ++i) {
         QPair<QString, QString> pair = queryItems.at(i);
-        if (pair.first == QL1S("location"))
-            subscriptionUrl = pair.second;
-        else if (pair.first == QL1S("title"))
-            subscriptionTitle = pair.second;
+        if (pair.first == QL1S("location")) {
+            QString temp = pair.second;
+            subscriptionUrl = QUrl(temp).path();
```
nowrep: That doesn't look right. What about `QUrl::fromPercentEncoding`?

innermous: That works correctly, not sure QUrl::fromPercentEncoding will.

nowrep: It works, but maybe the correct fix is just `QUrlQuery::queryItems(QUrl::FullyDecoded))`

innermous: Well, if it is, it's up to you then :)

nowrep: You can update the pull request if you want :)

innermous: I'd check correctness of that first, if you haven't yet.

innermous: You are right, that was so obvious.

nowrep: No, it wasn't. I needed to check how it is used in other place to see the error ;)

innermous: So update PR?

nowrep: Yes, please.

PR #1815, opened 2015-12-16
Fix downloadmanager displaying wrong filename

innermous:
Trying to save image file with custom name instead default one (generated from URL).
custom
Albeit file saves with custom file name, download manager prefers default one.
default

### DIFF
--- a/src/lib/downloads/downloadmanager.cpp
+++ b/src/lib/downloads/downloadmanager.cpp
@@ -256,7 +256,7 @@ void DownloadManager::download(QWebEngineDownloadItem *downloadItem)
 
     // Create download item
     QListWidgetItem* listItem = new QListWidgetItem(ui->list);
-    DownloadItem* downItem = new DownloadItem(listItem, downloadItem, QFileInfo(downloadPath).absolutePath(), fileName, openFile, this);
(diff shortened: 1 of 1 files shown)


src/lib/downloads/downloadmanager.cpp
```diff
@@ -256,7 +256,7 @@ void DownloadManager::download(QWebEngineDownloadItem *downloadItem)
 
     // Create download item
     QListWidgetItem* listItem = new QListWidgetItem(ui->list);
-    DownloadItem* downItem = new DownloadItem(listItem, downloadItem, QFileInfo(downloadPath).absolutePath(), fileName, openFile, this);
+    DownloadItem* downItem = new DownloadItem(listItem, downloadItem, QFileInfo(downloadPath).absolutePath(), QFileInfo(downloadPath).fileName(), openFile, this);
```
nowrep: `downloadPath` is always created from `fileName` so how does this helps?

innermous:
Download manager displays incorrect filename in download list, it's weird behavior IMO.

src/lib/downloads/downloadmanager.cpp
innermous: .fileName() extracts *correct* file name from path, whilst filename variable extracts file name from original URL. You can check that with qDebug().

PR #1818, opened 2015-12-18
Added feature to search Google for image

innermous:
This feature is really lacking one :)

### DIFF
--- a/src/lib/webengine/webview.cpp
+++ b/src/lib/webengine/webview.cpp
@@ -728,6 +728,12 @@ void WebView::createImageContextMenu(QMenu* menu, const WebHitTestResult &hitTes
     connect(act, SIGNAL(ctrlTriggered()), this, SLOT(userDefinedOpenUrlInNewTab()));
     menu->addAction(act);
     menu->addAction(QIcon::fromTheme("edit-copy"), tr("Copy image ad&dress"), this, SLOT(copyLinkToClipboard()))->setData(hitTest.imageUrl());
+    QUrl imageSearchUrl("https://www.google.com/searchbyimage?site=search&image_url=" + act->data().toString());
+    Action* gsact = new Action(tr("Search &Google for image"));
+    gsact->setData(imageSearchUrl);
+    connect(gsact, SIGNAL(triggered()), this, SLOT(openUrlInSelectedTab()));
+    connect(gsact, SIGNAL(ctrlTriggered()), this, SLOT(openUrlInBackgroundTab()));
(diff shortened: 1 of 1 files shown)


ke7ofi:
Is this something that should be built into the browser? It seems like something better suited to a addon or JS thing.

innermous:
I asked maintainer before PR, he said this sounds good. 
Can't see this as addon tho, this functionality is pretty common.

cranes-bill:
What does mean „_pretty common_“!? Firefox doesn't have this function. I can't say about other browsers for sure, but i think more of them also don't have it. Even there is a Chrome [add-on](https://chrome.google.com/webstore/detail/search-by-image-by-google/dajedkncpodkggklbegccjpmnglmnflm), so probably even Chrome don't have this as a commonly  function. Furthermore, some people might prefer different search engine.....

ke7ofi:
That's the other thing - keeping things engine-agnostic is good.

src/lib/webengine/webview.cpp
```diff
@@ -728,6 +728,12 @@ void WebView::createImageContextMenu(QMenu* menu, const WebHitTestResult &hitTes
     connect(act, SIGNAL(ctrlTriggered()), this, SLOT(userDefinedOpenUrlInNewTab()));
     menu->addAction(act);
     menu->addAction(QIcon::fromTheme("edit-copy"), tr("Copy image ad&dress"), this, SLOT(copyLinkToClipboard()))->setData(hitTest.imageUrl());
+    QUrl imageSearchUrl("https://www.google.com/searchbyimage?site=search&image_url=" + act->data().toString());
```
nowrep: Use `hitTest.imageUrl()`. Also, this may need to be percent encoded.

innermous:
All right, I think you are right, we need sort of browser add-on here.

nowrep:
I didn't say to close it.
I think it is useful feature, and even though it would be great if it were search provider agnostic, the fact is, that only google supports it. So there is not much else to do (same as the actions for translate).

cranes-bill:
@nowrep says: ... the fact is, that only google supports it.....
I'm not a quite agree with that.

ke7ofi:
@nowrep @cranes-bill There's TinEye, for example.

innermous:
@nowrep bing, yandex, and so on...
I thought, that it would be better if we make this feature plug-in, giving user a choice, whether he wanna use it at all, if so - what search engine is preferred... There's a little room for user options, but if u insist, I'll just edit PR and that will do.

PR #1820, opened 2015-12-20
Fix endless loading invalid webpages in speed-dial

### LINKED ISSUE #1763: Prevent invalid web-pages from loading endlessly in SpeedDial
Please try the following, reproducible steps:

(1) Go to SpeedDial page
(2) Add the following SpeedDial-Entry, which is a non-existing web-page

iq1

(3) Press `Apply`

(4) QupZilla will try to load this page endlessly and make the SpeedDial-Page flickering, which will lead to a very high system load.

Application version: 1.9.99 (a750023)
Qt version: 5.6.0
Platform: openSuSE Tumbleweed (Linux)

I did not test it with [qt-5.5] and [1.8] branch.

innermous:
This can be naive approach, but I believe *external.speedDial.changed(allPages())* call is just redundant. More than that, it screws everything, making endless loop when added URL is invalid.

Closes #1763 .

### DIFF
--- a/src/lib/data/html/speeddial.html
+++ b/src/lib/data/html/speeddial.html
@@ -287,8 +287,6 @@ function setTitleToUrl(url, title) {
         var span = box.getElementsByTagName('span')[0];
         span.innerText = title;
     }
-
-    external.speedDial.changed(allPages());
 }
 
 function setImageToUrl(url, img_source) {


nowrep:
But this won't propagate the change of speed dial title to all speed dial instances (in tabs).

PR #1834, opened 2016-01-14
Default to NoProxy rather than HttpProxy

kkofler:
The HttpProxy setting neither makes sense nor works without a concrete
proxy being set. So default to something that has any chance of working
without manual configuration. If you need an HTTP proxy, you will have
to configure it anyway.


nowrep:
Thanks

PR #1835, opened 2016-01-14
Make automatic searching from the address bar optional

kkofler:
Some users (like me) really don't like it when their typos get sent to a
search engine without asking.

### DIFF
--- a/src/lib/navigation/locationbar.cpp
+++ b/src/lib/navigation/locationbar.cpp
@@ -139,10 +139,13 @@ void LocationBar::setText(const QString &text)
 
 void LocationBar::updatePlaceHolderText()
 {
-    QString engineName = qzSettings->searchWithDefaultEngine ?
-                         mApp->searchEnginesManager()->defaultEngine().name :
-                         mApp->searchEnginesManager()->activeEngine().name;
-    setPlaceholderText(tr("Enter URL address or search on %1").arg(engineName));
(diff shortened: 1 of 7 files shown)


ke7ofi:
What's the default according to your revision?

kkofler:
The default is true (enabled), matching what the code previously did.

The followup patch I've just pushed fixes a crash caused by the new setting.

ke7ofi:
Cool, :+1: then.

nowrep:
Is that crash inside QtWebEngine?

In any case, it's better to do nothing than trying to load `my search` as url.

kkofler:
Your code does not crash. Trying to load the non-URL as a URL (i.e., what fcae1fd was doing) was what was crashing. a570beb drops the offending part of fcae1fd (exactly as you suggest: it does nothing rather than trying to load `my search` as a URL).

I tested fcae1fd + a570beb right after I pushed a570beb, and the feature works exactly as I promised: If I uncheck the automatic search option, then strings like `test` that are not valid URLs are no longer sent to the search engine (but simply ignored).

nowrep:
This is where it asserts when loading the incorrect url:

```
[0114/101447:FATAL:navigation_controller_impl.cc(927)] Check failed: active_entry->site_instance() == rfh->GetSiteInstance(). 
```

PR #1848, opened 2016-02-07
Allow to split shared data from prefix

berniyh:
My distribution uses a prefix that is different from /usr for cross-compile capability.
However, shared data should still go to /usr/share as expected. Thus qupzilla installs to the wrong directories.
The changes fix this, so that one can now define where the data should go explicitely. If not specified it should behave like previously.

At the same time I replaced some hardcoded paths.


nowrep:
Thanks

PR #1853, opened 2016-02-09
Feature: implemented bookmark editing on the fly

innermous:
Title stands for itself.


src/lib/bookmarks/bookmarkstools.h
```diff
@@ -82,6 +82,9 @@ class QUPZILLA_EXPORT BookmarksTools
     static bool addBookmarkDialog(QWidget* parent, const QUrl &url, const QString &title, BookmarkItem* folder = 0);
     static bool bookmarkAllTabsDialog(QWidget* parent, TabWidget* tabWidget, BookmarkItem* folder = 0);
 
+    // Edit Bookmark Dialog
+    static bool editBookmarkDialog(BrowserWindow* window, BookmarkItem* item);
```
nowrep: Make it just `QWidget *parent` instead of BrowserWindow.

PR #1914, opened 2016-04-03
added sh to OS X build instructions

r-a-v-a-s:
Without the `sh` I get `command not found`.


r-a-v-a-s:
I didn't use the `./` which apparently does the same thing.

PR #1919, opened 2016-04-04
ensured the window geometry is remembered on OS X

r-a-v-a-s:
It was always starting based on:
```
int windowWidth = desktop->availableGeometry().width() / 1.3;
int windowHeight = desktop->availableGeometry().height() / 1.3;
```
because quitApp isn’t called on OS X.

### DIFF
--- a/src/lib/app/browserwindow.cpp
+++ b/src/lib/app/browserwindow.cpp
@@ -1324,6 +1324,11 @@ void BrowserWindow::closeEvent(QCloseEvent* event)
     Settings settings;
     bool askOnClose = settings.value("Browser-Tabs-Settings/AskOnClosing", true).toBool();
 
+    settings.beginGroup("Browser-View-Settings");
+    settings.setValue("WindowMaximised", windowState().testFlag(Qt::WindowMaximized));
+    if (!isFullScreen()) settings.setValue("WindowGeometry", saveGeometry());
+    settings.endGroup();
+
     if (mApp->afterLaunch() == MainApplication::RestoreSession && mApp->windowCount() == 1) {
         askOnClose = false;
     }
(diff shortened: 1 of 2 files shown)


nowrep:
If quitApp isn't called on OS X, then saveSideBarWidth won't be called either.

What about moving the save settings part of `quitApp` function (basically the whole function except last `mApp->quitApplication()`), and call it from closeEvent()?

r-a-v-a-s:
Yes I had the same thought after submitting this.
I also noticed that quitApp always returns true.
So it seems like quitApp could be eliminated, and the logic could be reduced to:

```c++
#ifndef Q_OS_MAC
    if (mApp->windowCount() == 1) mApp->quitApplication();
#endif
```

Something else I'm wondering is if it's really necessary to call quit() in quitApplication.
Won't accepting the close event accomplish that?
I need to setup my PC for testing soon...

nowrep:
> Something else I'm wondering is if it's really necessary to call quit() in quitApplication.
> Won't accepting the close event accomplish that?

There may be other windows (popup windows, download manager, ..) so accepting close event is not enough.

r-a-v-a-s:
Oh right... ;-]

Do you want me to close this and submit another,
or just add another commit to this?

nowrep:
Add another commit.

r-a-v-a-s:
What do you think?

src/lib/app/browserwindow.cpp
```diff
@@ -1324,6 +1324,7 @@ void BrowserWindow::closeEvent(QCloseEvent* event)
     Settings settings;
     bool askOnClose = settings.value("Browser-Tabs-Settings/AskOnClosing", true).toBool();
 
+
```
nowrep: Remove

src/lib/app/browserwindow.cpp
```diff
@@ -1345,18 +1346,11 @@ void BrowserWindow::closeEvent(QCloseEvent* event)
...
-        else {
-            event->ignore();
-        }
+    saveSettings();
 
-        return;
-    }
-#endif
+    #ifndef Q_OS_MAC
+        if (mApp->windowCount() == 1) mApp->quitApplication();
```
nowrep:
```
if (mApp->windowCount() == 1)
    mApp->quitApplication();
```

src/lib/app/browserwindow.cpp
```diff
@@ -1399,15 +1393,9 @@ bool BrowserWindow::quitApp()
         settings.setValue("WebSearchBarWidth", m_navigationToolbar->splitter()->sizes().at(1));
         settings.setValue("SideBarWidth", m_sideBarWidth);
         settings.setValue("WebViewWidth", m_webViewWidth);
-
-        if (!isFullScreen()) {
-            settings.setValue("WindowGeometry", saveGeometry());
-        }
+        if (!isFullScreen()) settings.setValue("WindowGeometry", saveGeometry());
```
nowrep: Same here

r-a-v-a-s:
Jak si přejete. ;-]

nowrep:
Thanks!

Issue #1896, opened 2016-03-27
overlapping bookmarks on the bookmarks toolbar


r-a-v-a-s:
Version: 1.9.99
OS X 10.11.4 / retina display

<img width="446" alt="screen shot 2016-03-27 at 11 38 31 am" src="https://cloud.githubusercontent.com/assets/12013378/14067118/f118bab2-f411-11e5-8925-116afac2618e.png">

nowrep:
Same bug is also on non-retina screen.

ke7ofi:
I can't replicate it on Linux.

PR #1920, opened 2016-04-04
fixed overlapping bookmarks on the toolbar; issue #1896

r-a-v-a-s:
Letting Qt decide the spacing fixes the problem; and there appears to be 0 spacing anyhow.


nowrep:
Thanks, I hope it won't break it for other OSes.

PR #1958, opened 2016-04-25
Use `rev-parse` instead of `show-ref`

Martii:
* Dev builds currently use masters commit ref and makes it difficult to determine which version is really installed with About &rarr; Help. Use the current HEAD when generating the `Makefile`s for `DGIT_REVISION`

**NOTES**
* The bash scripts are somewhat duplicated with the unused `--short` parameter but omitted at this time in case a long hash is wanted.
* `make distclean` is still needed since those defines are hard-coded into those affected `Makefile`s

### DIFF
--- a/git_hooks/post-checkout
+++ b/git_hooks/post-checkout
@@ -2,4 +2,4 @@
 #
 # Writes git revision into git_revision file
 #
-git show-ref refs/heads/master | cut -d " " -f 1  > git_revision
+git rev-parse HEAD | cut -d " " -f 1  > git_revision
--- a/git_hooks/post-checkout
+++ b/git_hooks/post-checkout
@@ -2,4 +2,4 @@
 #
 # Writes git revision into git_revision file
 #
-git rev-parse HEAD | cut -d " " -f 1  > git_revision
+git rev-parse HEAD > git_revision
--- a/git_hooks/post-commit
+++ b/git_hooks/post-commit
@@ -2,4 +2,4 @@
 #
 # Writes git revision into git_revision file
 #
-git show-ref refs/heads/master | cut -d " " -f 1  > git_revision
+git rev-parse HEAD | cut -d " " -f 1  > git_revision
--- a/git_hooks/post-commit
+++ b/git_hooks/post-commit
@@ -2,4 +2,4 @@
 #
 # Writes git revision into git_revision file
 #
-git rev-parse HEAD | cut -d " " -f 1  > git_revision
+git rev-parse HEAD > git_revision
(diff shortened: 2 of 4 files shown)


git_hooks/post-checkout
```diff
@@ -2,4 +2,4 @@
 #
 # Writes git revision into git_revision file
 #
-git show-ref refs/heads/master | cut -d " " -f 1  > git_revision
+git rev-parse HEAD | cut -d " " -f 1  > git_revision
```
nowrep: `git rev-parse HEAD` returns only commit hash, so the `cut` is not needed anymore.

Martii: Just getting a feel for this PR... I already did that on this repo's checkout with a branch there... but... if someone has `git_revision` already on their machine it's going to present an issue.

btw when does Travis get finished typically?

nowrep: > if someone has git_revision already on their machine it's going to present an issue.

Why?

> btw when does Travis get finished typically?

10-15 minutes, but it is irrelevant for this pull request.

Martii: > Why?

Because I haven't fully *(our terminology here)* PR READY'd this yet... I am not sure why you didn't use `--short` in the first place and this `git` method since it's been in there for a really long time e.g. did you want it stripped out and just use `--short` or keep it as is like I put it in with your abstract? 

Travis build **failed** btw and it's quite relevant... I"m unfamiliar with travis so far *(only used mocha)*... but learning as I go along here.

nowrep: >  I am not sure why you didn't use --short in the first place

Because I just found one way to get the desired result and didn't care about the best way to do it.

> Travis build failed btw and it's quite relevant... I'm unfamiliar with travis so far (only used mocha)... but learning as I go along here.

No, just issue with Qt 5.7 PPA.

Martii: > didn't care about the best way to do it.

Getting a feel here still... so do you want to care now, or should I just remove the affected cuts with one left in in case someone has an old build installed? :question:  *(e.g. [here](https://github.com/Martii/qupzilla/blob/b1258085ad9d5febb6f6708d4eac19ce89698a15/scripts/getrevision.sh#L17) and [here](https://github.com/Martii/qupzilla/blob/b1258085ad9d5febb6f6708d4eac19ce89698a15/scripts/getrevision.sh#L8))*

nowrep:
Yep, looks fine now.
And also it's not in case someone has old build, the thing with `git_revision` is that I used to build from tarballs for my PPA and I updated the `git_revision` file there, so even though there were no `.git` directory, it still showed git revision in About dialog.

PR #1963, opened 2016-04-29
Remove redundant mkdir instruction

innermous:
Following mkdir is really redundant, you know.


Martii:
Presume you mean it's already done with [#1963/files#diff-`d3e9d29`](https://github.com/QupZilla/qupzilla/pull/1963/files#diff-d3e9d295c88b4168fe8ca4e632a33ab0R98). If so that would be a +1 here. :)

PR #1975, opened 2016-05-13
RSS

adl-88:
Hello guys,

where can add RSS links? I cannot find it.


nowrep:
Why do you open pull request?
RSS was removed in 2.0

Martii:
> Why do you open pull request?

There are some bot accounts running around creating bad PRs again... so it's probably just that.

PR #1979, opened 2016-05-16
Mac: Allow developer openssl configs

Martii:
- Homebrew is currently production state for openssl e.g. runtimes, snapshots, or whatever other synonym is used
- Add some ~~error~~ warning protection if neither are found.
- Precedence is production _(pro)_ then development _(dev)_
- Homebrew isn't always updated to current production release states. See #2085

Post followup for #1971

---

PR READY

### DIFF
--- a/src/lib/lib.pro
+++ b/src/lib/lib.pro
@@ -479,10 +479,36 @@ os2 {
 }
 
 mac {
-    # homebrew openssl
-    BREW_OPENSSL = $$system("brew --prefix openssl")
-    INCLUDEPATH += $$BREW_OPENSSL/include
-    LIBS += -L$$BREW_OPENSSL/lib
+    # openssl
+    HAS_OPENSSL = FALSE
+
+    # production openssl via homebrew
+    system(brew -v):HAS_BREW=TRUE
+    equals(HAS_BREW, TRUE) {
+      OPENSSL = $$system(brew --prefix openssl)
+      exists($$OPENSSL) {
+        HAS_OPENSSL = TRUE
+      } else {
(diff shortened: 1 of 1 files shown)


Martii:
This isn't quite ready `make distclean` failed.

nowrep:
Do we really want it? I think requiring `brew` is enough, after all it's de facto standard to have it on your Mac.

Martii:
> Do we really want it?

If anyone gets into developing/testing openssl and needs a test platform instead of the **prebuilt install binaries** then yes. This is a minor glitch as most people won't be deleting their openssl with brew and/or ssl linkage but I usually do forward and reverse regressions just to be sure. I'm still a bit green on .pro and haven't found a variable that tells me which target definition it's in. That's all I really need to remove the glitch... alternatively I could just not `errror` and go back to `warning`... but it will compile galore. If this was a plain Makefile I would modify to have only the first target definition run this e.g. `make` and `make all`... anything else would skip the test. Qt is a bit peculiar that it creates `clean` and `distclean` although it's nice that it does this... just need to determine which "mode" it's in... it's probably a simple one liner too but it's evading my searches... not to mention I need sleep as usual. ;) :)

Martii:
I'd have to modify/contribute to Qt directly in order to get some scopes in here that would allow termination of the build process... so just flipped it to `warning`.

Martii:
``` sh-session
$ brew upgrade openssl
...
Error: openssl 1.0.2k already installed
```

Ref:
* #2326

Martii:
Closing from lack of interest.

PR #2011, opened 2016-06-23
Fix position of SpeedDial items in RTL

SafaAlfulaij:
The position of the items in Speed Dial is mirrored as the float attribute wasn't changed according to the layout.


nowrep:
Thanks a79395b0fb3db5603d65e7b73f43450362484c16

PR #2019, opened 2016-06-30
Implement audio mute for webtabs

innermous:
This need some manual testing and tuning in the case something is not working smoothly/poor UX design.
If everything is alright, I hope the code is good too.


Martii:
Sample vid: https://www.youtube.com/watch?v=O70Ww9vzjvg

Qt 5.6.1 with ROSA Linux

``` sh-session
webtab/webtab.cpp: In constructor ‘WebTab::WebTab(BrowserWindow*)’:
webtab/webtab.cpp:145:33: error: ‘recentlyAudibleChanged’ is not a member of ‘QWebEnginePage’
     connect(m_webView->page(), &QWebEnginePage::recentlyAudibleChanged, tabIcon(), &TabIcon::updateAudioIcon);
                                 ^
webtab/webtab.cpp: In member function ‘bool WebTab::isMuted() const’:
webtab/webtab.cpp:298:31: error: ‘class WebPage’ has no member named ‘isAudioMuted’
     return m_webView->page()->isAudioMuted();
                               ^
webtab/webtab.cpp: In member function ‘void WebTab::setMuted(bool)’:
webtab/webtab.cpp:303:24: error: ‘class WebPage’ has no member named ‘setAudioMuted’
     m_webView->page()->setAudioMuted(muted);
                        ^
webtab/webtab.cpp: In member function ‘bool WebTab::isMuted() const’:
webtab/webtab.cpp:299:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
Makefile:4830: recipe for target '../../build/webtab.o' failed
make[1]: *** [../../build/webtab.o] Error 1
make[1]: Leaving directory '~/repo/git/QupZilla/martii/qupzilla/src/lib'
Makefile:44: recipe for target 'sub-src-lib-make_first' failed
make: *** [sub-src-lib-make_first] Error 2
```

Qt 5.7.0 OS X 10.11.5 *(El Capitan)*

Doesn't appear to work but did compile.

Which Qt and platform did you code with this PR?

nowrep:
This needs Qt 5.7 so it should have appropriate ifdefs

Martii:
Kind of figured that but thought I'd try anyhow. Looks like my distro distribution of Qt5 won't be of much help for a while.

Here's an OS X screen cap if it's useful:

ytwpr2019a

... don't see the icon mentioned anywhere in the tab. Will be neat if this gets implemented. :)

innermous:
@nowrep right, I forgot about it at all. Fixes coming.
@Martii I assume it's just ifdefs troubles.

innermous:
@Martii seems like it won't for <5.7.

mutetabs
This is how it looks for me, I noticed a little glitch with icon. Well, imo, such an icon doesn't match well and needs adjusting at least.

Martii:
OS X look:

ytwpr2019b

Tab 1: State paused
Tab 2: State playing
Tab 3: State muted *(if not muted muxed with other tab inputs)*

Fully functional although sitting on top of the favicon here. Kewl. :)

innermous:
@Martii plus, if it's muted, icon will remain no matter what until you explicitly unmute it.
I really need to do sth about icons, which means to adjust better and maybe find better fitting icon. No designer, I took icon I found at https://github.com/google/material-design-icons

Martii:
The Oxygen theme, or equivalent, might have some "speaker"s that are, perhaps GPL, redistributable too in Nix.

innermous:
@Martii if audio icon was separate from favicon(not as it's part) we could just use QStyle::SP_MediaVolumeMuted and etc. which are standard icons and which could be replaced by icon from user's theme.
But @nowrep preferes it as part of favicon, and I, frankly, even don't have an idea how to add it as separate icon to bar anyway.

Martii:
Hadn't thought of that... Keeping it as part of the favicon might look unusual especially if the favicon is a speaker or black coloring in nature *(unless Qt has an invert value mask?)*. 

Any way to make it appear on hover without too much eye distraction?

innermous:
So what's your thoughts on this matter guys? Should we leave this as is or need changes?
@nowrep did you test it?

cranes-bill:
@innermous,
This is very ugly to me. What will happen, if you make _mute_ icon with double width and first half will be transparent? Does this will looks as separate from favicon?

Sorry for the bad English!

src/lib/webtab/webtab.cpp
```diff
@@ -142,6 +142,9 @@ WebTab::WebTab(BrowserWindow* window)
     connect(m_webView, SIGNAL(loadStarted()), this, SLOT(loadStarted()));
     connect(m_webView, SIGNAL(loadFinished(bool)), this, SLOT(loadFinished()));
     connect(m_webView, SIGNAL(titleChanged(QString)), this, SLOT(titleChanged(QString)));
+#if QT_VERSION >= QT_VERSION_CHECK(5,7,0)
+    connect(m_webView->page(), &QWebEnginePage::recentlyAudibleChanged, tabIcon(), &TabIcon::updateAudioIcon);
```
nowrep: Only this needs to be #ifdef'd

src/lib/webtab/webtab.cpp
```diff
@@ -292,6 +295,24 @@ void WebTab::setPinned(bool state)
     m_isPinned = state;
 }
 
+#if QT_VERSION >= QT_VERSION_CHECK(5,7,0)
+bool WebTab::isMuted() const
+{
+    return m_webView->page()->isAudioMuted();
```
nowrep: And this

src/lib/webtab/webtab.cpp
```diff
@@ -292,6 +295,24 @@ void WebTab::setPinned(bool state)
...
+}
+
+void WebTab::setMuted(bool muted)
+{
+    m_webView->page()->setAudioMuted(muted);
```
nowrep: And this. So just #ifdef the code that cannot be built with Qt 5.7 (= just call to new QWebEnginePage functions/signals).

nowrep:
Well, it really needs better ui. So let's just show the mute icon next to favicon. In `TabIcon::updateAudioIcon` you would call let's say `resize(32, 16)` or `resize(16, 16)` respectively.

innermous:
@nowrep okay, I thought about it, and it sounds like good decision.

innermous:
@nowrep what if we stick to sound icon in the corner but give it a white border like in firefox? It looks quite good there.

nowrep:
@innermous Show screenshot please.

innermous:
@nowrep 
untitled

Martii:
> what if we stick to sound icon in the corner

That would be handy when multiple tabs are running and need a quicker toggle than going through every tab that has an active output sound.

> but give it a white border

This would ideally be a type of filter added to the image to do this programmatically, if possible, to give an edge effect mask... do those exist in Qt yet? ... Perhaps http://doc.qt.io/qt-5/qml-qtgraphicaleffects-glow.html or similar?

nowrep:
Looks better, can you move it to bottom right corner?

nowrep:
>  This would ideally be a type of filter added to the image to do this programmatically, if possible, to give an edge effect mask... do those exist in Qt yet? ... Perhaps http://doc.qt.io/qt-5/qml-qtgraphicaleffects-glow.html or similar?

Yes, something like that would be possible (the link you sent is QML class though). But if we "hardcode" the icon (eg. not use from desktop theme) then it is just easier to add the glow to image itself.

Martii:
> then it is just easier to add the glow to image itself.

OS X likes to do a lot of shading and restyling on its own is why this possibly won't be true... e.g. if it was coded we can balance the effects if needed between *nix/win and os x.

innermous:
I could use some help with styling, if possible.

Martii:
Part of a static save model, as @nowrep mentioned, is that the png already has transparency... and blending the "white border" into it will not be a simple process.

Here is a blown up GIMP view of the [audiomuted.png](https://github.com/innermous/qupzilla/blob/c2f08a6b2702a0334e33fcc0c1e5b5b531623084/src/lib/data/icons/other/audiomuted.png) showing the various alpha levels visually *(if you see the checkerboard pattern then it's partially transparent)*:
alpha opt

Using a programmatic blur/glow/feather mask e.g. load the [QPixmap](http://doc.qt.io/qt-5/qpixmap.html), possibly convert it to a [QImage](http://doc.qt.io/qt-5/qimage.html), resize just slightly bigger, applying a filter, then "resaving" it in a memory buffer to display is the usual process for "sprite" based manipulations... filters can come in all flavors but what is available in Qt base is probably what would be the easiest to manage.

I still like the idea of a separate icon *(similar to the new tab icon)* but overlaying can be a bit more complicated.

nowrep:
Why did you close it?

Martii:
@innermous 
Please reopen... besides some UI artifacts it's quite useful... as @nowrep said we can work on prettyfying the UI up a bit later. It's going to take me some time to get used to 3 different platforms and developing... and I'm mostly in observation mode here with a few 2 cents but you do nice work. :)

innermous:
@nowrep I just don't know how to implement this. You can merge the feature itself without tracking icon and switching audio state via context menu and hopefully someone will provide nice UI solution for that.
In best case we should do something with tab bar itself and show icon after the title(and of course some workaround for pinned tabs). But I don't know how to implement that yet. Neither can I implement what was Suggested before with white borders.

Martii:
Let me see if I can anti-alias the audiomute icon... this will take some time as pixel art takes lots of time to get it right. I had an idea today where I might be able to do it quickly but depends on how it looks... I do wish OS X and Qt was a little more consistent with WYSIWG as compared to the other os'.

cranes-bill:
What about none-transparent background, just completely over favicon? Here is white with black border for exanple:

Martii:
@cranes-bill 
That is a similar effect... trying a round circle with a feather but that's not much better than the base transparent icon over the favicon. 

1. Not sure if @nowrep wants color or not... have to pay attention to color blind people usually with software... that's mostly why most icons are black and white.
2. Good idea but the favicon is absent with a full square unfortunately there.
3. Another issue that I might look into is why OS X is flipping the close and favicon *(with sound mod applied here)* ... but may not be able to do anything about that unless it's in Qt itself... base toolkit widgets in OS X are usually handled by Apple.

I decided to load Fx and see how they did it... they did an anti-alias **and** a near white border with select pixels... looks about the same as the anti-aliased one I tried here for Qupzilla.

Going to tweak with no "sound bars" eminating for mute as that probably will look better... the muted icon is the one that renders the worse.

Martii:
So it looks like OS X is keeping the icon at a maximum of 16x16 and clipping quite a bit... tried it in the upper right corner to see how much was getting clipped and it's about 4 pixels on the right side.

It also appears from @innermous screen cap it is also clipping at 18x18 on that platform.

In short... the audiomuted and audioplaying icons would need to be approximately 8x8 to fit on both platforms in the "pinned" state... that's **really tiny** to make anything discernible especially with a white border. e.g. I don't think that an overlay is going to be useful. The code is solid in this PR but the UI itself isn't WYSIWG e.g. separating it into a separate tab or "status bar" icon would be preferable I think.

So some choice summaries:
* Leave the PR UI as is and work on it later... not pretty but definitely functional so far.
* Create another tab like entry with all of the available options.
* Create a status bar icon that would simulate the tab like entry.
* Unpinned I would recommend splitting the icon completely out of the favicon just like Fx did... but that only solves half the UI issue.
* Obliterate the favicon altogether in favor of a sound toggle icon *(not really recommended although looks nice)*

While I'm not a graphic artist I did try everything in my arsenal including pixel inversion, white out, circular/square feathering, resizing *(smaller)*, etc. and nothing really looks good with the UI. Until the favicon bounding box is a true WYSIWG 16x16 pixel box for all platforms there is an issue on resizing to 16x32 with a pixel or two left/right margin.

@nowrep 

Per [here](https://github.com/QupZilla/qupzilla/pull/2019#issuecomment-230433637) I did a brief search on these methods for resizing when you posted that and couldn't find them... e.g. what class is that based off of in Qt? e.g. a doc reference please.

nowrep:
Let's fix the code, merge it and modify the UI later. Please update the ifdefs only where needed.

innermous:
And leave the current icons or remove them?

nowrep:
Leave the current icons

innermous:
@nowrep I don't get it.
I thought, I ifdef'd everything that should be ifdef'd. Or you mean, there's some code that could be compiled without ifdefs?

Martii:
@innermous 
I believe he means put the bare minimum`#if QT_VERSION >= QT_VERSION_CHECK(5,7,0)` in that is necessary to compile in Qt 5.6.1 **and** Qt 5.7.0+ in order to get this merged. e.g. I presume all the platforms will eventually not be supported with Qt 5.6.1 down the line and the project will move forward... it would be less lines to remove. It's just a matter of preference I think... me personally I would leave any affected code that requires Qt 5.7.0 in but that's just me.

I'll be the first to pipe in if my distro *(or any other major distro)* moves from Qt 5.6.1 to Qt 5.7.x . KDE/Plasma in this distro is still Qt 4 stable mostly but they added in an additional Qt 5 unstable sequence.

Since there really isn't a full policy on migration versions for QupZilla it's an "on the fly" sort of thing.

innermous:
IMO everything that's related to stuff not supported on 5.6.1  (i.e. this mute audio code itself) should be ifdef'd, even if it compiles... Correct me if I am wrong.

nowrep:
No, I prefer to have as few ifdefs as possible.

src/lib/webtab/webtab.cpp
```diff
@@ -292,6 +295,26 @@ void WebTab::setPinned(bool state)
     m_isPinned = state;
 }
 
+bool WebTab::isMuted() const
```
innermous: One small thing, static analyzer issues about reaching the end of non-void function, when you are building with Qt <5.7.0
I assume, it's not a big deal, since eventually all code will be adjusted to newer versions of Qt?

innermous:
Oh snap, and unused parameter is treated as error.

src/lib/webtab/webtab.cpp
nowrep: Return false when < 5.7.0

innermous: I did it already, man :)

src/lib/tabwidget/tabicon.cpp
```diff
@@ -35,6 +36,10 @@ TabIcon::TabIcon(QWidget* parent)
     m_animationPixmap = QIcon(QSL(":icons/other/loading.png")).pixmap(288, 16);
     m_framesCount = m_animationPixmap.width() / m_animationPixmap.height();
 
+    m_audioIconDisplayed = false;
```
nowrep: Move it to constructor initializer list

src/lib/tabwidget/tabicon.cpp
```diff
@@ -113,4 +128,26 @@ void TabIcon::paintEvent(QPaintEvent* event)
         p.drawPixmap(r, m_animationPixmap, QRect(m_currentFrame * pixmapSize, 0, pixmapSize, pixmapSize));
     else
         p.drawPixmap(r, m_sitePixmap);
+
+#if QT_VERSION >= QT_VERSION_CHECK(5,7,0)
```
nowrep: This probably doesn't need ifdef? m_audioIconDisplayed should not ever be true on Qt < 5.7.0?

src/lib/tabwidget/tabicon.cpp
```diff
@@ -113,4 +128,26 @@ void TabIcon::paintEvent(QPaintEvent* event)
...
+        else
+            p.drawPixmap(QPointF(width() * 0.25, 0), m_audioPlayingPixmap);
+    }
+#endif
+}
+
+#if QT_VERSION >= QT_VERSION_CHECK(5,7,0)
+void TabIcon::mousePressEvent(QMouseEvent *event)
+{
+    qreal x = event->localPos().x();
```
nowrep: #if QT_VERSION >= QT_VERSION_CHECK(5,7,0)

src/lib/tabwidget/tabicon.cpp
```diff
@@ -113,4 +128,26 @@ void TabIcon::paintEvent(QPaintEvent* event)
...
+    qreal y = event->localPos().y();
+    // if audio icon is clicked - we don't propagate mouse press to the tab
+    if (m_audioIconDisplayed && x >= width() * 0.25 && y < height() * 0.75)
+        m_tab->toggleMuted();
+    else
```
nowrep: #endif

nowrep:
Please fix the last comments and I'll merge it. Thanks!

src/lib/tabwidget/tabicon.cpp
innermous: I don't get this comment.
This method should only override if version is >= 5.7.0

src/lib/tabwidget/tabicon.cpp
innermous: Didn't get this either...

innermous:
Updated patch set.

Martii:
@innermous 
Thanks. :) Compiles under OS X and simple retesting. Nice work.

innermous:
@Martii still requires a lot of redesigning to make this look better.

cranes-bill:
@innermous, can you look at [VIVALDI](https://vivaldi.com/) source? I know it's not a Qt browser, but there _**the mute**_ is replacing _**the favicon**_, so it would be nice if this part of the code can be implemented here!

Issue #1943, opened 2016-04-13
Black pages in v2.0.0


hooddy:
Hi there
I have black pages in new v2.0.0 in any pages like start, new, any internet pages etc.
Please advice me
![untitled](https://cloud.githubusercontent.com/assets/6899176/14512711/80100f02-01f2-11e6-934b-f912915ca264.png)

Martii:
I get this through RDC/RDP with 2.0.0 and current HEAD builds. Didn't used to do this with 1.8.9.

vatterspun:
Confirmed.  John Haller over at PortableApps is reporting serious issues with QupZilla 2.0.0 and 2.0.1:
http://www.portablefreeware.com/forums/viewtopic.php?p=81621#p81621 ... they are keeping their version at 1.8.9: http://portableapps.com/apps/internet/qupzilla-portable

Martii:
Re:

> ... 2.0.1 don't work on my Windows XP ... virtual machines.

Popup dialog showing of:

```
qupzilla.exe - Entry Point Not Found
The procedure entry point strnlen could not be located in the dynamic link library msvcrt.dll
```

Clicked okay about a dozen or so times and it finally launched qupzilla. No black screens here in this clean VM. msvcrt.dll possibly could use an update *(or worse a downgrade depending on how qupzilla is built and what binaries are distributed, if any)*

briskycat:
I get a bunch of the same "strnlen Entry Point Not Found" messages at startup with the version 2.0.2 on WinXP SP3. After that everything seems to work just fine.

Chaz6:
Running QupZilla from the Fedora 25 KDE Live DVD results in a black screen when browsing.

Kurisutian:
Same here on Manjaro-ARM (latest) running on a RPi3. I'm running version 2.0.2 and it just shows a black screen for less than a minute and crashes afterwards.
Is there any known fix? I'd love to run qupzilla there since it's fast and less resource hungry than the other browsers.

Martii:
Don't recall when but RDC/RDP hasn't done this to me for a while.

PR #2027, opened 2016-07-10
GM WIN: Add CR to detection

Martii:
Saving in Notepad++ with Windows line ending prevents script injection

Post fix for #1964 and some vague references at https://github.com/QupZilla/qupzilla/commit/ce67c7a455a58b3cf64491054688e4c9e9593c3a

PR NEEDS TESTING under Windows before it's ready. This is the standard GM fix which appears to be needed in Qt string handling as well... @nowrep please compile it and test under Windows before merging. *(OS X and Linux distros appear good to go)*

---

NOTE: In plain WIN7 NOTEPAD.EXE it will show up visibly as a normal looking .user.js however when NIX/OSX line endings are present it will be all on one line.

---

PR READY confirmation [here](https://github.com/QupZilla/qupzilla/commit/ce67c7a455a58b3cf64491054688e4c9e9593c3a#commitcomment-18196952) from pejakm

### DIFF
--- a/src/plugins/GreaseMonkey/gm_script.cpp
+++ b/src/plugins/GreaseMonkey/gm_script.cpp
@@ -238,7 +238,7 @@ void GM_Script::parseScript()
 
     const QString fileData = QString::fromUtf8(file.readAll());
 
-    QzRegExp rx(QSL("(?:^|[\\r\\n])// ==UserScript==(.*)(?:[\\r\\n]|\\r\\n)// ==/UserScript==(?:[\\r\\n]|$)"));
+    QzRegExp rx(QSL("(?:^|[\\r\\n])// ==UserScript==(.*)(?:\\r\\n|[\\r\\n])// ==/UserScript==(?:[\\r\\n]|$)"));
     rx.indexIn(fileData);
     QString metadataBlock = rx.cap(1).trimmed();
 
@@ -248,7 +248,7 @@ void GM_Script::parseScript()
     }
 
     QStringList requireList;
-    QzRegExp rxNL(QSL("(?:[\\r\\n]|\\r\\n)"));
+    QzRegExp rxNL(QSL("(?:\\r\\n|[\\r\\n])"));
 
     const QStringList lines = metadataBlock.split(rxNL, QString::SkipEmptyParts);
     foreach (QString line, lines) {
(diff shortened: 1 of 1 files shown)


nowrep:
What about to use `\R` instead?

Martii:
> What about to use `\R` instead?

Unfamiliar with that string token... documentation reference?

Ref:
* [QRegExp - Characters and Abbreviations for Sets of Characters](http://doc.qt.io/qt-5.7/qregexp.html) *(no `id` or `name` to bookmark with this link... so please scroll down)*
* [QString](http://doc.qt.io/qt-5/qstring.html)

Martii:
Not entirely sure if QRegularExpression handles that PERL Unicode Regular Expression item... doc refs don't  seem to mention it.

Ref:
* http://perldoc.perl.org/perlre.html#Regular-Expression#Character-Classes-and-other-Special-Escapes

Martii:
I would venture to say that the `\R` is a bit buggy... once I put in a UTF-8 BOM check *(another thing for the next step... nodes default string handling is UTF-16 so the characters are bit different with regular expression tests on OUJS but same concept)*... `\R` fails to load BOM scripts altogether e.g. `\r` re's correctly whereas `\R` does not. *(Qt 5.6.1)*

This regular expression implementation is a bit unusual as well... `\uXXXX` has to be done for Unicode character identification whereas control characters have to be `\\r` and `\\n` respectively... that isn't all that consistent  I think unless I'm missing something major.

Martii:
@pejakm 

Would you retest this under Windows with the cumulative commits please... there was a preexisting glitch with new installs which should be fixed now. e.g. `@name` should be displayed in the UI instead of the entire UserScript metadata block.

This also covers older UTF-8 editors out there that may use CR+LF

pejakm:
@Martii I'm sorry, I don't have Windows installed, nor do I have disk resources (or time) to even try such thing.

nowrep:
Why do we need windows to test it? Ideally this should be unit tested.

Martii:
@pejakm 
Thanks... then you, or your source, probably has an editor that is doing Windows line endings then.

@nowrep 
Just confirming build... don't have to if you don't want to. I installed Community Visual Studio 2015 on the Windows box, along with Qt and developer OpenSSL... but since I haven't built under Windows yet with all the component hooks and #1943 exists which makes it more difficult to test as I have to go into the server tower room which is not exactly comfortable.

> should be unit tested.

Which is why I did Hello, World U.T.s. [referenced above](https://github.com/QupZilla/qupzilla/pull/2027#issuecomment-232504552). :)

PR #2052, opened 2016-08-08
Show Bookmark Toolbar when Location Bar gets focus

loimu:
Optional behaviour providing quick access to bookmarks without wasting the vertical space

Resembles old Opera's style of accessing the bookmarks with the "Quick Bar"

### DIFF
--- a/src/lib/preferences/preferences.cpp
+++ b/src/lib/preferences/preferences.cpp
@@ -217,7 +217,7 @@ Preferences::Preferences(BrowserWindow* window)
     // NOTE: instantBookmarksToolbar and showBookmarksToolbar cannot be both enabled at the same time
     ui->instantBookmarksToolbar->setChecked(settings.value("instantBookmarksToolbar", false).toBool());
     ui->showBookmarksToolbar->setChecked(settings.value("showBookmarksToolbar", true).toBool());
-    ui->instantBookmarksToolbar->setDisabled(settings.value("showBookmarksToolbar").toBool());
+    ui->instantBookmarksToolbar->setDisabled(settings.value("showBookmarksToolbar", true).toBool());
     ui->showBookmarksToolbar->setDisabled(settings.value("instantBookmarksToolbar").toBool());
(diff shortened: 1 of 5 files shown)


src/lib/preferences/preferences.cpp
```diff
@@ -214,7 +214,13 @@ Preferences::Preferences(BrowserWindow* window)
     //APPEREANCE
     settings.beginGroup("Browser-View-Settings");
     ui->showStatusbar->setChecked(settings.value("showStatusBar", false).toBool());
+    // NOTE: instantBookmarksToolbar and showBookmarksToolbar cannot be both enabled at the same time
```
nowrep: This should be taken care of instead of just note in code.

Martii:
Cool description of the feature... this would be quite useful when fully functional. I'll run it against OS X in the next few days.

innermous:
I like this :+1:

src/lib/preferences/preferences.cpp
Martii: @nowrep 
Actual OS X testing and logic flow shows that this should be okay... unless you want to force a focus event on browser startup as a "reminder".

Martii: Here's an alternate mockup of the same UI logic but uses more screen real estate and gets rid of the cool name of "instant bookmarks toolbar":
alternate

innermous: @Martii why not just make it drop down combobox with 3 options (display, display on focus, do not display) ?

Martii: @innermous 
Because it's a mockup of screencaps and not Code ... harder to redraw. ;) That could work too. That would remove the CON of loss of screen real estate for sure.

innermous: I can implement settings code and post a patch for that here.

nowrep:
Merged bc47a9aba5cba58a2dfe2d8345bb84d19fc47f33

PR #2055, opened 2016-08-11
Show navigation in new tab in fullscreen mode

innermous:
I think it might be a handy enhancement.

### DIFF
--- a/src/lib/tabwidget/tabwidget.cpp
+++ b/src/lib/tabwidget/tabwidget.cpp
@@ -230,6 +230,16 @@ void TabWidget::updateClosedTabsButton()
     m_buttonClosedTabs->setEnabled(canRestoreTab());
 }
 
+bool TabWidget::isCurrentTabFresh() const
+{
+    return m_currentTabFresh;
+}
+
+void TabWidget::setCurrentTabFresh(bool currentTabFresh)
+{
+    m_currentTabFresh = currentTabFresh;
+}
+
 void TabWidget::tabBarOverFlowChanged(bool overflowed)
 {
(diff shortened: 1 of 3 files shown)


src/lib/tabwidget/tabwidget.cpp
```diff
@@ -695,6 +695,11 @@ bool TabWidget::canRestoreTab() const
     return m_closedTabsManager->isClosedTabAvailable();
 }
 
+bool TabWidget::isCurrentTabNew()
+{
+    return m_urlOnNewTab == weTab()->url();
```
nowrep: Umm, no. This is just wrong, you will treat *any* current tab with loaded m_urlOnNewTab as new tab.

innermous: Okay, I'll think about better way to do it.

src/lib/tabwidget/tabwidget.cpp
```diff
@@ -367,6 +378,10 @@ int TabWidget::addView(const LoadRequest &req, const QString &title, const Qz::N
     connect(webTab->webView(), SIGNAL(wantsCloseTab(int)), this, SLOT(closeTab(int)));
     connect(webTab->webView(), SIGNAL(urlChanged(QUrl)), this, SIGNAL(changed()));
     connect(webTab->webView(), SIGNAL(ipChanged(QString)), m_window->ipLabel(), SLOT(setText(QString)));
+    connect(webTab->webView(), &WebView::urlChanged, this, [this](const QUrl& url) {
```
nowrep: `const QUrl &url`

Issue #2049, opened 2016-08-05
Closing right/left tabs (Enhancement)


tsujan:
It would be nice if QupZilla had `Close Tabs to the left` and `Close Tabs to the right` in addition to `Close Other Tabs` on the tab right-click menu. I use QupZilla on a regular basis instead of Firefox and think those menu-items would be handy.

Thanks for this great browser!

Edit: My bad! I wrote `CloneZilla` instead of `QupZilla `!!

tsujan:
Thanks!

PR #2062, opened 2016-08-19
Issue #2049 close tabs to right & close tabs to left

std46:


src/lib/tabwidget/tabbar.cpp
```diff
@@ -137,6 +138,24 @@ void TabBar::closeAllButCurrent()
     }
 }
 
+void TabBar::closeToRight(){
```
nowrep: Coding style - brace on new line

```
void TabBar::closeToRight()
{
    // ...
```

Same in other places.

std46:
fixed formatting

PR #2068, opened 2016-08-27
Add initial printing support with QtWebEngine >= 5.7.0

kkofler:
This works by printing to PDF first, then sending the PDF to a printer if printing to a physical printer was requested.

On *nix (including Mac OS X), we use the lpr command, which is typically provided by CUPS nowadays. Currently, no options beyond the printer name are passed, there is room for improvement there. (Okular has a FilePrinter class that handles this in a more sophisticated way.)

On Windows, we use ShellExecuteW with the "printto" verb. In that case, the printer name is unfortunately the only thing that can be passed. The user experience may also be suboptimal depending on the PDF viewer installed on the system. (E.g., Adobe Reader is reported to ignore SW_HIDE.) And if there is no PDF viewer installed at all, it will of course not work at all. But it is the best we can do without bundling something like GhostScript (or requiring the user to install it).

I successfully tested the revision I am now submitting on Fedora Rawhide (actually, I backported the patches to QupZilla 2.0.1 and tested that, but that should not make any difference in practice): it can definitely bring something to paper (and printing to PDF also works), honoring more options is TODO as explained above. I cannot test the Windows code.

### DIFF
--- a/src/lib/app/browserwindow.cpp
+++ b/src/lib/app/browserwindow.cpp
@@ -636,6 +636,7 @@ void BrowserWindow::printPage()
             weView()->page()->printToPdf(dialog->printer()->outputFileName(), dialog->printer()->pageLayout());
         } else {
             QTemporaryFile tempFile(QDir::tempPath() + QSL("/QupZillaPrintXXXXXX.pdf"));
+            tempFile.setAutoRemove(false);
             if (tempFile.open()) { // create the file name, and the file to reserve it
                 tempFile.close(); // close, but don't delete the file
                 weView()->page()->printToPdf(tempFile.fileName(), dialog->printer()->pageLayout());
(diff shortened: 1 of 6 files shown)


The-Compiler:
Note that real printing support will be [added](https://wiki.qt.io/New_Features_in_Qt_5.8) in Qt 5.8 too.

kkofler:
Support on the QupZilla end will be needed there too. In fact, I would add the code for Qt 5.8 as an additional `#if` on top of this changeset.

Also note that the way the code in Qt 5.8 will work is not all that different from my implementation: It also starts by printing to a PDF. What it does then is different though: instead of running `lpr` or `ShellExecuteW`, it uses the PDFium bundled with Chromium (which is not built in Qt 5.7, but will be built in Qt 5.8 exactly for this purpose) to render the PDF to images and sends those to the QPrinter.

At least compared to `lpr`, this is actually an inferior technique: CUPS `lpr` can usually convert from PDF directly to the printer's vector language rather than to a bitmap. On the other hand, you get the option handling from QPrinter for free, whereas my `lpr` invocation is currently missing the options. (There is existing code for that in Okular though, which could be ported.) So I am not sure whether it wouldn't make sense to keep `lpr` on *nix even with Qt 5.8. It probably depends on whether the option handling for `lpr` gets implemented.

On Windows, my `ShellExecuteW` approach is clearly a hack, and the Qt 5.8 approach is definitely better there. All the code I found for printing PDF files on Windows is either the same `ShellExecuteW` hack or does what Qt 5.8 does (i.e., uses a PDF renderer to render to a bitmap and prints that). There is really no clean way to do any better on Windows: to do any better, you'd have to ship the whole CUPS filter chain, somehow identify the proper CUPS driver for the printer (if there is even one) and then send the RAW file to the printer using Windows APIs.

But in any case, since Qt 5.8 is not even out yet, and since I don't think QupZilla will want to require Qt 5.8 immediately (e.g., it still supports 5.6 now that 5.7 is out), I think having Qt 5.7 support in cannot hurt.

nowrep:
Looks good, but I would prefer to (at least in QProcess case) run external processes asynchronously - so it doesn't block if it takes too much time or something goes wrong.

kkofler:
Well, `lpr` asynchronously queues the print request, so it normally returns immediately.

kkofler:
That said, I can convert it to an actual QProcess object and the finished signal, the only drawback is that the code will be more complicated. (I don't want to use the fire&forget QProcess::startDetached because I want to delete the temporary file if `lpr` cannot be run for some reason.)

kkofler:
Distribution packagers interested in this patch will probably also want: https://codereview.qt-project.org/170632

Without that fix, margin settings are not honored by QtWebEngine 5.7.0. (It is already fixed in 5.8 as part of a larger commit.) There is not much we can do about that in QupZilla. We could pass margins to `lpr` (at least if it is the CUPS `lpr`), but then those would be added in addition to the ones already in the PDF (i.e., the resulting margins would be too big), and the content would be scaled down instead of reflown.

nowrep:
Thanks!

PR #2099, opened 2016-10-15
Fix information about OS version

solodyagin:
Please check your mt.exe path in main.pro


nowrep:
What is the issue that this pull request fixes?

nowrep:
This should be fixed with d9202d12868da03bd2c5a7f10c98ea04b720499c - Win 8, 8.1 and 10 detection

r-a-v-a-s:
Does this do more than [QSysInfo::prettyProductName()](http://doc.qt.io/qt-5/qsysinfo.html#prettyProductName)?

Example: OS X El Capitan (10.11)

nowrep:
Alright, so we need the manifest thing. But the rest of the patch is unnecessary big, I don't really care for every possible Windows version as QupZilla will not ever run on the majority of them.
The same with Linux distributions, it doesn't matter which distribution is QupZilla running on. Just that it is running on Linux is enough.

Can you please update the patch with only fix for the Windows 8+ detection, and remove everything else? Thanks.

Martii:
Seems like the suggestion from @r-a-v-a-s might be less of a footprint change and less to maintain. Some of this information is useful to testers although welcome to Windows with 50 million sub-versions. Perhaps some of the missing OS types might be suitable upstream at Qt?

PR #2112, opened 2016-11-03
Update languages.nsh

hakking:
Add "Hong Kong Cantonese" (LangId 3076, zh-HK).


nowrep:
New languages also needs to be added to `installer.nsi`

PR #2170, opened 2017-01-21
Update the cache size settings

ThaFireDragonOfDeath:
Changes:
Exchanged the QSlider with a QSpinBox.
Added a MB suffix in the spin box (looks like: "50 MB" in the box).
Enhanced the cache size range from 20MB - 200MB to 1MB - 999999999MB.
Removed the current cache size from the label before the spin box.
Removed some unnecessary frames around the maximum label to get the Label on the correct place.

Sense:
Gives the user more control over the cache size.

Note:
The cache limit doesn't work (in the old and my version), but that is another problem.

Screenshot:
qupzilla-newcache_ex


nowrep:
Thanks

PR #2173, opened 2017-01-22
Added option to delete all cookies on exiting

ThaFireDragonOfDeath:
Changes:
Add new option in the cookie manager: "Remove cookies on exit".

Sense:
Gives the user more control over their privacy.

Notes:
New string "Remove cookies on exit" needs to be translated.
German String: "Cookies beim Beenden löschen"
Same feature for the cache would be nice.

Screenshot:
screenshot-cookiemanager


nowrep:
Can you please rename the settings key as it is in `v1.8`? https://github.com/QupZilla/qupzilla/blob/v1.8/src/lib/cookies/cookiejar.cpp#L54

ThaFireDragonOfDeath:
No problem, wait a minute...

Issue #2180, opened 2017-01-24
OS X: Some menu icons missing


Martii:
I've been away doing other things but slowly getting back into the swing of things here.

The issue is in OS X current head of 24b6db866a7239655d796ae0c53a3ce58ac1a61d

With some of the conversions of PNGs to SVGs they seem to be missing in certain views. This is best shown by the below highlighted image. The VNC viewer is OS X and the image below is the VM of openSUSE *(yes there was a "method to the madness" in getting a platform where I could compile the current HEAD for comparison)*.

<img src="https://cloud.githubusercontent.com/assets/114709/22248245/eb8d8a82-e1fa-11e6-83ec-c42bba1866b7.png" width="102" height="64" title="missing icons"/>

nowrep:
You must include Qt Svg Icon Image Plugin in the bundle.

Martii:
Well if you expect me to do that for distribution then you'll have to walk me through it... the whole Qt 5.7.1 was installed on OS X so it makes it valid. Thanks. I am not using the online installer under OS X since it's basically a blank slate and that's for Linux.

nowrep:
It should be done by `macdeployqt` that is invoked with `mac/macdeploy.sh` script. I'm not sure where exactly the plugins will be, but it is somewhere inside the bundle. So you should take a look and make sure it's there.

In Qt installation they are in `$QTDIR/plugins/` and you specifically want `$QTDIR/plugins/imageformats/libqsvg.dylib`

Martii:
This is the folder dump:

```
$ pwd
/Users/user/Qt/Qt5.7.1/5.7/clang_64/plugins
$ ls
audio/           geoservices/     playlistformats/ sensorgestures/
bearer/          iconengines/     position/        sensors/
canbus/          imageformats/    printsupport/    sqldrivers/
designer/        mediaservice/    qmltooling/
generic/         platforms/       sceneparsers/

$ cd imageformats
$ ls
libqdds.dylib*               libqmacjp2_debug.dylib*
libqdds.dylib.dSYM/          libqmacjp2_debug.dylib.dSYM/
libqdds_debug.dylib*         libqsvg.dylib*
libqdds_debug.dylib.dSYM/    libqsvg.dylib.dSYM/
libqgif.dylib*               libqsvg_debug.dylib*
libqgif.dylib.dSYM/          libqsvg_debug.dylib.dSYM/
libqgif_debug.dylib*         libqtga.dylib*
[...]

Issue #456, opened 2012-07-14
german translation improvement when asked to store password


ff2000:
qupzilla asks me
"Soll QupZilla das Passwort für ... auf ... speichern?"
But there is no button "Ja", "Speichern", or so, just "Erinnern". That's somewhat misleading. When I was asked this for the first time, I was unsure if this will immediately store ("speichern") the password or just queue it somewhere and on close there will be a popup, that will remind me ("Erinnern") on the still-to-store passwords and if qupzilla should preceed. (I am quite oldfashioned and don't know ANYTHING about what's state-of-the-art ATM ;))
If you think about (and try ;)) it it's clear, but in the first place you are irritated.

P.S.:
I asked myself, if i should write this bug in german, as it only concerns the german translation. Can I do so for the next time?

PR #2184, opened 2017-01-26
Mac: Copy missing dylib to bundle and document

Martii:
* This fixes the missing svg icons in the menu drop-down. This is a valid issue to be addressed.

NOTE: Be sure to set the `QTDIR` environment variable to point to your platform Qt installation... **EDIT**: ... if you answer `y`es to the prompt at any time is when it is needed.

Applies to #2180 and GH labels should be modified accordingly to accommodate this Qt feature.

### DIFF
--- a/mac/macdeploy.sh
+++ b/mac/macdeploy.sh
@@ -8,6 +8,7 @@
 MACDEPLOYQT="macdeployqt"
 LIBRARY_NAME="libQupZilla.2.dylib"
 PLUGINS="QupZilla.app/Contents/Resources/plugins"
+QTPLUGINS="QupZilla.app/Contents/PlugIns"
 
 if [ -n "$1" ]; then
  MACDEPLOYQT=$1
@@ -48,7 +49,7 @@ if echo "$answer" | grep -iq "^y"; then
 
     FILE="$QTDIR/plugins/iconengines/libqsvgicon.dylib"
     if [ -f "$FILE" ]; then
-      cp $FILE $PLUGINS
+      cp $FILE $QTPLUGINS
     else
       echo "$FILE: No such file"
       exit 1
@@ -58,7 +59,7 @@ if echo "$answer" | grep -iq "^y"; then
 else
   printf '\nChecking for prior deploy image library plugins at target...\n'
 
-  rm $PLUGINS/libqsvgicon.dylib
+  rm $QTPLUGINS/libqsvgicon.dylib
 fi
 
 # run macdeployqt
--- a/mac/macdeploy.sh
+++ b/mac/macdeploy.sh
@@ -45,7 +45,9 @@ if echo "$answer" | grep -iq "^y"; then
     printf '\nPlease set the environment variable for the Qt platform folder.\n\texample:\n\t$ export QTDIR="$HOME/Qt/5.7/clang_64"\n'
     exit 1
   else
-    printf '\nCopying known, missing, Qt native library plugins to target...\n'
+    printf '\nCopying known, missing, Qt native library plugins to target bundle...\n'
+
+    mkdir -p $QTPLUGINS
 
     FILE="$QTDIR/plugins/iconengines/libqsvgicon.dylib"
     if [ -f "$FILE" ]; then
@@ -57,9 +59,9 @@ if echo "$answer" | grep -iq "^y"; then
 
   fi
 else
-  printf '\nChecking for prior deploy image library plugins at target...\n'
+  printf '\nChecking for prior deploy image Qt native library plugins at target bundle...\n'
 
-  rm $QTPLUGINS/libqsvgicon.dylib
+  rm -Rf $QTPLUGINS
 fi
 
 # run macdeployqt
--- a/mac/macdeploy.sh
+++ b/mac/macdeploy.sh
@@ -55,3 +55,4 @@ $MACDEPLOYQT QupZilla.app
 # create final dmg image
 cd ../mac
 ./create_dmg.sh
+
--- a/mac/macdeploy.sh
+++ b/mac/macdeploy.sh
@@ -51,7 +51,7 @@ if echo "$answer" | grep -iq "^y"; then
 
     FILE="$QTDIR/plugins/iconengines/libqsvgicon.dylib"
     if [ -f "$FILE" ]; then
-      cp $FILE $QTPLUGINS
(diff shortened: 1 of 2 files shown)


Martii:
Unless there are any suggestions this is PR READY to resolve the issue with the QupZilla OS X mac build/deploy script.

nowrep:
Please remove the $QTDIR from doc and just show error message from script when it is not specified.

nowrep:
Also you can query libraries directory with `qmake -query QT_INSTALL_LIBS`

mac/macdeploy.sh
```diff
@@ -7,6 +7,7 @@
 
 MACDEPLOYQT="macdeployqt"
 LIBRARY_NAME="libQupZilla.2.dylib"
+PLUGINS="QupZilla.app/Contents/Resources/plugins"
```
nowrep: Is this really the correct path? I think this is "our" Resources path - eg. for application. Libraries (Frameworks) should be in different path.

mac/macdeploy.sh
```diff
@@ -19,23 +20,50 @@ cd bin
...
+# copy all QupZilla plugins into bundle
+test -d $PLUGINS || mkdir $PLUGINS
+cp plugins/*.dylib $PLUGINS/
 
 # fix libQupZilla
 install_name_tool -change $LIBRARY_NAME @executable_path/$LIBRARY_NAME QupZilla.app/Contents/MacOS/QupZilla
 
 # fix plugins
-for plugin in QupZilla.app/Contents/Resources/plugins/*.dylib
+for plugin in $PLUGINS/*.dylib
```
nowrep: This should not be done for Qt plugins.

mac/macdeploy.sh
Martii: Doesn't work any place else with `install_name_tool`... even if I add an id to it it never finds it and conversely the executable of `QupZilla` nor the lib that QupZilla creates... tries `rpath` too. This is the only location it works at.

mac/macdeploy.sh
Martii: Do you mean the abstraction of the variable or the `install_name_tool`?

nowrep: I mean running the `install_name_tool`.

mac/macdeploy.sh
nowrep: Hmm, can you please upload somewhere the `QupZilla.app` after running macdeployqt on it?

Martii:
@nowrep
> Also you can query libraries directory with qmake -query QT_INSTALL_LIBS

Issue with that is with multiple versions of Qt installed... which one to pick. Luckily OS X doesn't have the same issue as Tumbleweed crashing when it is set in the `PATH` but if someone specifies a specific qmake the script wouldn't know which one to call and it could return the incorrect path. Hopefully no one on OS X is using `qmake-qt5` aliases either. So `qmake` usage in the script shouldn't be used.

mac/macdeploy.sh
Martii: Well that wouldn't be me then... all I did was abstract it... but according to the Qt docs custom Libraries should be set so they "announce themselves" for use.

mac/macdeploy.sh
Martii: Give me a moment...

Martii: Btw it will be a dmg... as my u/l speed isn't good with 152MiB folder vs the dmg which is 50.2MiB... just mount the image and then go to /Volumes/QupZilla or install it and then it should be r/w...  uploading now to dune... will send a link in email when it's finished. You'll also need a refresher on `otool -L`.

Martii: Mail sent.

mac/macdeploy.sh
Martii: > This should not be done for Qt plugins.

Rereading this a few times to digest a possible singular meaning... I didn't set that for the Qt plugin *(the libqsvgicon.dylib)* e.g. it's a flat out copy *(does work if I do set it though but decided it probably wouldn't be the right thing to do)*... so there is no issue here.  Read further down in the shell script.

mac/macdeploy.sh
Martii: @nowrep 



Okay here's what was tried last week and one tid bit today. Please refer to the image attached for the item numbers.

1. Qt source path for `libqsvgicon.dylib` is at `.../clang_64/plugins/iconengines/`
2. Target goal *where it should be* but *is not*... for this illustration it was created manually there but also programmatically with `macdeploy.sh`.  This does **not** work here even with `install_name_tool`.
3. Todays retest... it does work here and committed with fe9e41788dd08df673b48627ade7c5fac5b0d16b

A. The issue itself... when the dylib plugin is placed in a path where it is known to be loaded with `macdeployqt.sh` then it does finally load the "New Tab" and "New Window" svg icons.

I've also left the non-highlighted areas visible to show how Qt is laid out structure wise both pre-deploy and post-deploy of QupZilla.

nowrep: I find it strange that other plugins are correctly loaded from subdirectories, but `iconengines` not ...

**Docs**
During development, the directory for plugins is QTDIR/plugins (where QTDIR is the directory where Qt is 
installed), with each type of plugin in a subdirectory for that type, for example, styles. If you want your applications 
to use plugins and you don't want to use the standard plugins path, have your installation process determine the 
path you want to use for the plugins, and save the path, for example, by using QSettings, for the application to read when it runs. The application can then call QCoreApplication::addLibraryPath() with this path and your plugins will 
be available to the application. *Note that the final part of the path (for example, styles) cannot be changed.*

But if it works, then let's go with it.

Martii:
Followup for this with:

* d6b142bba2d9c73aff1201cacefaffc0a7423376
* Qt 5.9.1 snapshot *(`#456`)*
* OS X 10.12.5 *(Sierra)*
* Xcode 8.3.3
* Referenced Qt bug report from https://github.com/QupZilla/qupzilla/issues/2267#issuecomment-288191307 and https://bugreports.qt.io/browse/QTBUG-59609
* qupzilla 2184 followup
* Status **CONFIRMED FIXED**

@nowrep 
So let us know when this patch can be backed out for which QupZilla release... still need to wait until Qt 5.9.1 is in release too but the snapshot is good to go here with a manual backout. Cc: @srazi

Issue #2208, opened 2017-02-10
Mac: `macdeploy.sh` errors with Qt 5.8


Martii:
Re: 13cba7a069f208dd1b03b3426911a40f882d7bf2

The following happens with `macdeploy.sh` when attempting to create the dmg with Qt 5.8 *(not present with Qt 5.7.1)*:

``` sh-session
$ mac/macdeploy.sh
Do you wish to redistribute known, missing, Qt Library plugins (y/n)? 
Copying known, missing, Qt native library plugins to target bundle...
WARNING: Plugin "libqsqlodbc.dylib" uses private API and is not Mac App store compliant.
WARNING: Plugin "libqsqlpsql.dylib" uses private API and is not Mac App store compliant.
ERROR: no file at "/opt/local/lib/mysql55/mysql/libmysqlclient.18.dylib"
ERROR: no file at "/usr/local/lib/libpq.5.dylib"
We just want to make sure it's not in use
hdiutil: detach failed - No such file or directory
Creating writable DMG from template
Reading Driver Descriptor Map (DDM : 0)…
Reading Apple (Apple_partition_map : 1)…
Reading disk image (Apple_HFS : 2)…
..
Reading  (Apple_Free : 3)…
Elapsed Time:  1.525s
Speed: 98.3Mbytes/sec
Savings: 0.0%
created: /tmp/tmp-qupzilla.dmg
Attach DMG for write
/dev/disk2          	Apple_partition_scheme         	
/dev/disk2s1        	Apple_partition_map            	
/dev/disk2s2        	Apple_HFS                      	/private/tmp/tmp-qupzilla-release
Copying…
cp: /tmp/tmp-qupzilla-release/QupZilla.app/Contents/Resources/locale/cs_CZ.qm: No space left on device
cp: /tmp/tmp-qupzilla-release/QupZilla.app/Contents/Resources/locale/da_DK.qm: No space left on device
cp: /tmp/tmp-qupzilla-release/QupZilla.app/Contents/Resources/locale/de_DE.qm: No space left on device
cp: /tmp/tmp-qupzilla-release/QupZilla.app/Contents/Resources/locale/el_GR.qm: No space left on device
cp: /tmp/tmp-qupzilla-release/QupZilla.app/Contents/Resources/locale/es_ES.qm: No space left on device
cp: /tmp/tmp-qupzilla-release/QupZilla.app/Contents/Resources/locale/es_MX.qm: No space left on device
[...]

PR #2226, opened 2017-02-18
Mac: Refactor OS X dmg creation and misc bug fixes

Martii:
Had a little bit of extra time today to retest and refine... unless there are any different suggestions this should be PR READY. Please note I won't be able to make major mods until next week one day. Can be picked at until I get back if wished.

Here's how it looks now:
* how it looks

Applies to #2208


mac/macdeploy.sh
```diff
@@ -42,7 +42,7 @@ answer=$( while ! head -c 1 | grep -i '[yn]'; do true; done )
 stty $old_stty_cfg
 if echo "$answer" | grep -iq "^y"; then
   if [ -z ${QTDIR+x} ]; then
-    printf '\nPlease set the environment variable for the Qt platform folder.\n\texample:\n\t$ export QTDIR="$HOME/Qt/5.7/clang_64"\n'
+    printf '\nPlease set the environment variable for the Qt platform folder.\n\texample:\n\t$ export QTDIR="$HOME/Qt/5.8/clang_64"\n'
```
nowrep: I plan to call the `macdeploy.sh` from `qmake` so that we don't need to manually set these env variables (as qmake already knows the correct path).

mac/create_dmg.sh
```diff
@@ -1,26 +1,61 @@
...
 
-echo "Creating writable DMG from template"
-hdiutil convert QupZilla-template.dmg -format UDRW -o /tmp/tmp-qupzilla.dmg
+# NOTE: Value must currently match -volname exactly or an error happens in AppleScript with Finder
+VOLUME_TEMPLATE=QupZilla
 
-echo "Attach DMG for write"
-hdiutil attach /tmp/tmp-qupzilla.dmg -mountpoint /tmp/tmp-qupzilla-release
+echo "Ensuring working disk image template is not in use…"
+hdiutil detach $TMP/$WORK_TEMPLATE
```
Martii: Arggh... missed some quotes here.

PR #2244, opened 2017-03-09
Fixed new tab not opening on mouse mid button click

### LINKED ISSUE #2238: middle click does not work
The middle click does not work reliably to open links in new tabs. It can take 10 clicks or more before the input is registered. In Firefox on the same machine it works fine.

I'm using Qupzilla 2.1.1 on openSUSE Tumbleweed

joseriosneto:
Whenever the user pressed the middle button the mouse press handler
in WebPage issued an 'execJavaScript' which has its own event loop
causing the mouse release handler to be called before the field
'm_clickedUrl' get the url in the first handler. Since this field
is used to guard the logic in the second handler the new tab wasn't
opening.

Closes #2238

### DIFF
--- a/src/lib/webengine/webpage.cpp
+++ b/src/lib/webengine/webpage.cpp
@@ -141,7 +141,7 @@ QVariant WebPage::execJavaScript(const QString &scriptSource, quint32 worldId, i
         }
     });
 
-    loop->exec();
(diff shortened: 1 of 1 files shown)


nowrep:
Thanks

PR #2245, opened 2017-03-10
CI for Windows and macOS

srazi:
This patch add OSX build/deploy using Travis and also add integration for AppVeyor to support build/deploy under Windows. It will deploy on new tags (macOS DMG file, x86/x64 installers and installer for portable version).

To test deployed binaries [see](https://github.com/srazi/qupzilla/releases/tag/test_deploy_win_osx) (I have not tested DMG file)
 
Before merging you have to replace github api token by your tokens or disable deploy part.

**Windows Portable Installer:**
It is an installer that doesn't need admin rights to run and won't write to registry and also won't create an uninstaller..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ before_install:
     fi
 
 install:
+  - export QZ_DIR=$PWD
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
          sudo apt-get -qq update;
          sudo apt-get -qq install libssl-dev pkg-config libxcb-util0-dev;
@@ -30,14 +31,21 @@ install:
     fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
          if [[ "$QT" == "qt58" ]]; then
+              QT_VER=5.8;
+              export QTDIR="$HOME/Qt/$QT_VER/clang_64";
+              echo "Calling brew update...";
               brew update;
               brew outdated openssl || brew upgrade openssl;
-              brew install qt5;
+              QT_INSTALLER_FILE_NAME=qt-opensource-mac-x64-clang-$QT_VER.0;
+              wget "http://download.qt.io/official_releases/qt/$QT_VER/$QT_VER.0/$QT_INSTALLER_FILE_NAME.dmg";
+              hdiutil attach -noverify $QT_INSTALLER_FILE_NAME.dmg;
+              QT_INSTALLER=/Volumes/$QT_INSTALLER_FILE_NAME/$QT_INSTALLER_FILE_NAME.app/Contents/MacOS/$QT_INSTALLER_FILE_NAME;
+              echo "Silently installing Qt...";
+              $QT_INSTALLER --script $QZ_DIR/mac/qt-mac-silent-install.qs;
          fi
     fi
 
 script:
-  - export QZ_DIR=$PWD
   - QMAKE="qmake"
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
          CXXFLAGS="-Wextra -Werror";
@@ -47,9 +55,8 @@ script:
          MACDEPLOYQT="macdeployqt";
          CXXFLAGS="-Wextra -Werror -Wno-inconsistent-missing-override -Wno-deprecated-declarations";
          if [[ "$QT" == "qt58" ]]; then
-              export QTDIR=/usr/local/opt/qt5;
-              MACDEPLOYQT=/usr/local/opt/qt5/bin/macdeployqt;
-              QMAKE=/usr/local/opt/qt5/bin/qmake;
+              MACDEPLOYQT=$QTDIR/bin/macdeployqt;
+              QMAKE=$QTDIR/bin/qmake;
          fi
     fi
   - $QMAKE QMAKE_CXXFLAGS+="$CXXFLAGS"
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,21 +3,75 @@ dist: trusty
 language: cpp
 cache: ccache
 
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8.2
+
 env:
   - QT=qt58
 
 before_install:
-  - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
+  - export QZ_VER=2.1.99
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+         export DISPLAY=:99.0;
+         sh -e /etc/init.d/xvfb start;
+    fi
 
 install:
-  - sudo apt-get -qq update
-  - sudo apt-get -qq install libssl-dev pkg-config libxcb-util0-dev
-  - if [[ "$QT" == "qt58" ]]; then sudo apt-add-repository -y ppa:beineri/opt-qt58-trusty; sudo apt-get update -qq; sudo apt-get install -qq qt58tools qt58script qt58webengine qt58webchannel qt58declarative qt58x11extras; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+         sudo apt-get -qq update;
+         sudo apt-get -qq install libssl-dev pkg-config libxcb-util0-dev;
+         if [[ "$QT" == "qt58" ]]; then
+              sudo apt-add-repository -y ppa:beineri/opt-qt58-trusty;
+              sudo apt-get update -qq;
+              sudo apt-get install -qq qt58tools qt58script qt58webengine qt58webchannel qt58declarative qt58x11extras;
+         fi
+    fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+         if [[ "$QT" == "qt58" ]]; then
+              brew update;
+              brew outdated openssl || brew upgrade openssl;
+              brew install qt5;
+         fi
+    fi
 
 script:
+  - export QZ_DIR=$PWD
   - QMAKE="qmake"
-  - if [[ "$QT" == "qt58" ]]; then QMAKE="/opt/qt58/bin/qmake"; fi
-  - $QMAKE QMAKE_CXXFLAGS+="-Wextra -Werror"
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+         CXXFLAGS="-Wextra -Werror";
+         if [[ "$QT" == "qt58" ]]; then QMAKE="/opt/qt58/bin/qmake"; fi
+    fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+         MACDEPLOYQT="macdeployqt";
+         CXXFLAGS="-Wextra -Werror -Wno-inconsistent-missing-override -Wno-deprecated-declarations";
+         if [[ "$QT" == "qt58" ]]; then
+              export QTDIR=/usr/local/opt/qt5;
+              MACDEPLOYQT=/usr/local/opt/qt5/bin/macdeployqt;
+              QMAKE=/usr/local/opt/qt5/bin/qmake;
+         fi
+    fi
+  - $QMAKE QMAKE_CXXFLAGS+="$CXXFLAGS"
   - make -j2 || exit 1
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+         cp -v bin/libQupZilla*.dylib tests/autotests;
+    fi
   - cd scripts && ./run_tests.sh $QMAKE
+
+before_deploy:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+         cd $QZ_DIR;
+         ./mac/macdeploy.sh $MACDEPLOYQT;
+         mv bin/QupZilla.dmg bin/QupZilla-${QZ_VER}.dmg;
+    fi
+
+deploy:
+  provider: releases
+  api_key:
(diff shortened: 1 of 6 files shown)


mac/macdeploy.sh
```diff
@@ -34,12 +40,22 @@ do
...
+else
+   if [[ "$REDISTRIBUTE" == "NOQT" ]]; then
+     echo "$REDISTRIBUTE_QUESTION No"
+     answer="n"
+   else
+     # prompt and optionally copy additional Qt native plugin(s) into bundle
+     echo -n $REDISTRIBUTE_QUESTION
+     old_stty_cfg=$(stty -g)
+     stty raw -echo
+     answer=$( while ! head -c 1 | grep -i '[yn]'; do true; done )
```
srazi: @Martii Maybe I didn't get your point correct. But,
1- 2nd parameter is optional no changes needed for other codes using `macdeploy.sh`

2- If you ask me, that question is useless and should be removed completely. Since `macdeploy.sh` is mainly used by developers and maintainers and I think they always want to bundle Qt libraries.

srazi: The parameter `[<QT|NOQT>]` is used in `.travis.yml`

srazi:
> ... so I would hope that you would be open to suggestions instead of throwing subtle, proverbial, arrows at my question.

I am open to suggestions. Maybe my bad English causes you think I am not. I am sorry for my bad English.

> `./macdeploy.sh [<full-path-to-macdeployqt> [<QT|NOQT>]]`
... to indicate that exact parameter switch placement is needed. e.g. right now it states $1 is optional as well as $2

I agree.

> Adding an extra parameter to the shell script would make a more complicated issue vs. a calculated timeout.

tbh, my answers didn't mean that I am against this option. But as I don't have access to a Mac it takes some times to commit a fix. So I will wait for @nowrep to decide and then if he will  prefer another option I will fix it.

cranes-bill:
Hay, @srazi, glad to „see“ :eyes: you again here! I was thinking you have quit the project.

nowrep:
`macdeploy.sh` is useless without bundling Qt libraries, so please remove that option.

Martii:
@srazi 

I appreciate the consideration. My apologies if I came across anything but nice or misunderstood. :smile:

If you know how to use the `Qupzilla.pro` syntax to forward out the correct `$QTPLUGINS` from using `qmake` or `/opt/Qt.../qmake` that would be great to resolve the need altogether. In the short term I can just manually remove the known, missing, Qt libraries before a release to see if the file copy routine is no longer needed. I did my best to keep everything native Qt, then Shell Script, and the very least AppleScript since I'm one of the few people here that have a full blown Macintosh available for the project. I'm still reacquainting myself with Qt options so I do not know how to do what @nowrep asks with that just yet. His bread crumbs are helpful for me but sometimes I just can't find the solutions and have to defer to anothers expertise.

nowrep:
> If you know how to use the Qupzilla.pro syntax to forward out the correct $QTPLUGINS from using 

Something like this should work:

```
    macdeploysh.target = bundle
    macdeploysh.commands = scripts/macdeploy.sh $$[QT_INSTALL_BINS]/macdeployqt
    QMAKE_EXTRA_TARGETS += macdeployqt
```

and then it can be run with `make bundle`.

Martii:
> and then it can be run with `make bundle`.

@nowrep 

So in the shell script it would need to be parent path stripped from `$1` perhaps with `sed` in order to set `$QTPLUGINS` since the `bin` folder is different than the `plugins` folder... and then utilizing a relative file URI

So as I'm understanding with the shell script side...
1) Require the `$1` switch parameter
2) Take it's value and strip off `qmake` at the tail end of the argument.
3) Potentially append a relative file URI of `../plugins` and then do the file copying.
4) In the future rather than exec the shell script directly just use `make bundle` for OS X releases.

So for this PR it would probably need to do the same or similar thing since it's calling macdeploy.sh as well. @srazi, @nowrep is Travis able to query the current Qt being used with `qmake`?

srazi:
@cranes-bill  Hi thanks, No, I love QupZilla. I am just very busy with my university research and study.

@Martii 
> So for this PR it would probably need to do the same or similar thing since it's calling macdeploy.sh as well. @srazi, @nowrep is Travis able to query the current Qt being used with qmake? ❓

Yes. See the output of `qmake -query` and for example `qmake -query QT_VERSION`

Martii:
@srazi 

> Yes. See the output of `qmake -query` and for example `qmake -query QT_VERSION`

See https://github.com/QupZilla/qupzilla/pull/2184#issuecomment-275639110

Depending on where Travis is run from... either GH or locally... it could choose the wrong path especially when `qmake`'s bin path isn't in the `PATH` environment variable as needed under Tumbleweed  with a VM since it will hose the VM... forcing anyone using it to revert to a prior snapshot.

srazi:
> See #2184 (comment)

But we are using hard-coded qmake path in travis.yml.

> Can Travis just call make bundle and return the same tests? ❓

What did you mean with "the same tests"?

Martii:
> What did you mean with "the same tests"?

Testers lingo... generally means code paths... or logic testing. Just looking for the same potential issue that may crop up if you change `macdeploy.sh` to do [the above logic](https://github.com/QupZilla/qupzilla/pull/2245#issuecomment-286207117).

> But we are using hard-coded qmake path in travis.yml.

Understand... but if you look at https://github.com/QupZilla/qupzilla/commit/a3d598da0b8b8df8463798e0ac80162ea3874a2b there are multiple versions of Qt running through the iterations **if I'm understanding this correctly**. So it seems to me that when it runs through one version the libs included may be the correct path but then the other version if the paths are hard coded *(how this is done is the potential issue to look for I think)*. I haven't gone much beyond logic testing this locally as I don't even have Travis installed here.

Anyhow if you are comfortable on the proposed logic change I will go into observation mode since my experience with Travis isn't as good as yours. :smile: If there are any suggestions that I have with `macdeploy.sh` I'll pop in when I have some time later.

> To test deployed binaries [see](https://github.com/srazi/qupzilla/releases/tag/test_deploy_win_osx) (I have not tested DMG file)

One note on a potential request I saw here... I downloaded that .dmg and installed it... however when I ran QupZilla it was quite corrupt. e.g. never even got started and asked me to send a report off to Apple. So something is awry.

Martii:
Misc note I am guessing that GitHub *(GH)* doesn't apply the same file size limitation with Travis that it applies to individual users. 

I am curious if you did something else special to get such large files [under releases](https://github.com/srazi/qupzilla/releases/tag/test_deploy_win_osx)? :question: 

The last time I tried that GH barked at me saying it was too huge so I had to use my private server to distribute to @nowrep the .dmg file.

srazi:
> ... but if you look at a3d598d there are multiple versions of Qt running through the...

Indeed, each one of them will run on a separate worker (separate job that means a fresh os image that our travis script will install the right Qt before starting compile)

> One note on a potential request I saw here... I downloaded that .dmg and installed it...

Thanks for testing. When I checked the output I saw [these lines](https://travis-ci.org/srazi/qupzilla/jobs/209633577#L2525) but they are just some warnings. So maybe you can check what is the difference between dmg in my fork and the working dmg released by you?

Or maybe issue is with xcode and OS X version?
> Travis CI uses OS X 10.11.6 (and Xcode 7.3.1) by default .

See [Travis OS-X-Version](https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version)

> I am curious if you did something else special to get such large files under releases?

Nothing special. And also I uploaded a 52MB file to github release manually, without problem.

Martii:
> So maybe you can check what is the difference between dmg in my fork and the working dmg released by you?

Snipped report:
``` sh-session
Termination Reason:    DYLD, [0x1] Library missing

Application Specific Information:
dyld: launch, loading dependent libraries

Dyld Error Message:
  Library not loaded: /usr/local/Cellar/qt5/5.8.0_1/lib/QtWebEngineCore.framework/Versions/5/QtWebEngineCore
  Referenced from: /Users/user/Desktop/QupZilla.app/Contents/Frameworks/QtWebEngineWidgets.framework/Versions/5/QtWebEngineWidgets
  Reason: image not found
```

... this tells me that most likely the host used Homebrew to install Qt instead of the Qt Unified Installer... and that the `qmake` origins are incorrect. Framework error is saying the support DLL isn't combined into the bundle and looking for it outside of the bundle. So possible `install_name_tool` issue with the bundle option in Travis/Homebrew.

This is exactly what I'm concerned with Travis and it's subordinate calls not knowing the right path... also possible it's just the host machines configuration being foo'd.


> > Travis CI uses OS X 10.11.6 (and Xcode 7.3.1) by default .

Releases here are usually latest for maximum security and inter-op-ability. So:
* OS X 10.12.3
* Xcode 8.2.1
* Qt 5.8 and latest via the MaintenanceTool for build *(3rd octet but within QupZilla.pro specs)*
* Any known, missing, dylibs that `macdeployqt` forgot to include from corresponding version.
* openSSL from Homebrew latest which is 1.0.2k for latest security

Usually when @nowrep releases an actual release I'll attempt to update everything before the build. So Travis may want to be configured to use a newer "flag" for the default if it works out such as `osx_image: xcode8.2` to mirror release tested... but as I asked elsewhere in some misc. notes in another issue, I am not sure what the bottom end of QupZilla compatibility is with OS X. Ideally it would be nice to have a range but my production/deployment Mac is primarily for QupZilla at this time... so I keep it up to date.


> Nothing special...

Hmm maybe GH changed it again. Thanks. I'll give it another try next round.

srazi:
@Martii Would you please test [this one](https://github.com/srazi/qupzilla/releases/download/test_osx_11_12/QupZilla-2.1.99.dmg)?
`Changes:` It uses Xcode 8.2 and macOS 10.12 and openssl 1.0.2k

Martii:
> Would you please test [this one]?

Sure... same issue [as before](https://github.com/QupZilla/qupzilla/pull/2245#issuecomment-286607854) with the missing QtWebEngineCore. Installs but doesn't run and wants me to run off to Apple with that report.

srazi:
> Sure... same issue as before with the missing QtWebEngineCore. Installs but doesn't run and wants me to run off to Apple with that report.

Thanks. So we have to apply `install_name_tool` manually on `QtWebEngineWidgets` or maybe use: https://github.com/aurelien-rainone/macdeployqtfix

Martii:
> So we have to apply `install_name_tool` manually on `QtWebEngineWidgets` 

And probably the next Qt primary dep, and the next one, and so on.


> maybe use: https://github.com/aurelien-rainone/macdeployqtfix

Seems like a bad hack to work-around the issue of Homebrew not doing something correctly with the paths to start. But if you can isolate that concept to just Travis it would possibly help to merge this PR.

As far as development goes locally the Qt Unified Installer *(and even their static packages from Qt)* is preferred for multiple versions and multiple projects. The Homebrew openSSL "mostly does it right" albeit they drop the dylib in the incorrect folder *(but it still works)*... so perhaps that project can apply the same logic from that brew package to their Qt5.

Martii:
Another random thought I had yesterday is if the Qt installers could be abstracted to run without human interaction for Travis instead of using the boo'ged Homebrew version.

nowrep:
> if the Qt installers could be abstracted to run without human interaction

Yes, it can be done with Qt installer scripting.

srazi:
Well, I will use official Qt. But it is 1.3 GB vs Homebrew Qt 82MB. How Homebrew Qt is that small ?!

Martii:
> How Homebrew Qt is that small ?!

Well it's Homebrews packaging system so it may be gzipped, or equivalent, then they modify it so it's distributed via brew. It's quite possible they don't have the full Qt compliment available and just a stripped down Qt5. 

I usually turn off sources with the Unified Installer and that cuts down the size/time quite a bit. That's why scripting the Qt Unified Installer should take that into consideration... also there is another platform, ARM or Android, if I recall by default without looking at it... that could probably be dismissed.

srazi:
I added a commit to use official Qt for osx builds on travis. Also I did some cleanups and re-based branch on current master.

nowrep:
I'm ready to merge it, as soon as the bundle runs fine.

srazi:
> I'm ready to merge it, as soon as the bundle runs fine.

Good. I have tested it but better to wait for a double check by  @Martii

There are also some issues that I will create a new issue with a check list for them.

Martii:
@srazi 

>the new [DMG](https://github.com/srazi/qupzilla/releases/download/test_new/QupZilla-2.1.99.dmg) works correctly... ... for a double check by @Martii

+1 ... .dmg seems a little bigger than what I create for deploy but can figure that part out later. It's approximately 54.1MiB around where it should be about 51.5MiB.

@nowrep 

> ready to merge

+1

.travis.yml
```diff
@@ -3,21 +3,82 @@ dist: trusty
...
   - make -j2 || exit 1
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+         cp -v bin/libQupZilla*.dylib tests/autotests;
+    fi
   - cd scripts && ./run_tests.sh $QMAKE
+
+before_deploy:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+         cd $QZ_DIR;
+         ./mac/macdeploy.sh $MACDEPLOYQT;
```
Martii: NOTING: When `macdeploy.sh` and `QupZilla.pro` gets changed assume this is the only spot in that needs to be fixered here. *(sort of a question too)* :smile:

srazi:
> ... .dmg seems a little bigger than what I create for deploy but can figure that part out later. It's approximately 54.1MiB. Around where it should be is about 51.5MiB for v2.1.2ish.

No. The only difference are `libpq` and `libssl` that are together ~500 KB. It seems you compare value in different OSes 54.1MB ~= 51.5MiB

nowrep:
https://travis-ci.org/QupZilla/qupzilla/jobs/212240422

Martii:
Probably because I'm observing the 2.1.2 release I emailed @nowrep yesterday and yours is on `master`.... but I'll check on same system... closer... your DMG is ~51.6 MiB on same system.

srazi:
> https://travis-ci.org/QupZilla/qupzilla/jobs/212240422

Strange... was successfully compiled as PR and also in my fork.

srazi:
@nowrep I added QupZilla/qupzilla to appveyor, I don't know if there is a difference that you as owner add the project or me as a member.
Also it seems when merging you have not updated github api token in deploy section for both travis and appveyor.

srazi:
@nowrep Well, the commit 9af542aa03402eaa8b595b356871ae61c47522c7 didn't trigger an appveyor build so you should add it to appveyor,

PR #2255, opened 2017-03-16
Readme

Chocobo1:
* no need to warp URLs
* use https
* add alt attribute
* clone over secure channel


Chocobo1:
In action: https://github.com/Chocobo1/qupzilla/tree/readme#qupzilla-web-browser

PR #2263, opened 2017-03-19
TravisCI: utilize ccache on osx

Chocobo1:
* TravisCI: utilize ccache on osx
* TravisCI: show ccache stats

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,14 @@ install:
          fi
     fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+         echo "Calling brew update...";
+         brew update;
+         brew install ccache;
+         PATH="/usr/local/opt/ccache/libexec:$PATH";
+
          if [[ "$QT" == "qt58" ]]; then
               QT_VER=5.8;
               export QTDIR="$HOME/Qt/$QT_VER/clang_64";
-              echo "Calling brew update...";
-              brew update;
               brew outdated openssl || brew upgrade openssl;
               QT_INSTALLER_FILE_NAME=qt-opensource-mac-x64-clang-$QT_VER.0;
               wget "http://download.qt.io/official_releases/qt/$QT_VER/$QT_VER.0/$QT_INSTALLER_FILE_NAME.dmg";
@@ -60,7 +63,7 @@ script:
          fi
     fi
   - $QMAKE QMAKE_CXXFLAGS+="$CXXFLAGS"
-  - make -j2 || exit 1
+  - make -j2 CC=$CC CXX=$CXX || exit 1
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
          cp -v bin/libQupZilla*.dylib tests/autotests;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,7 @@ script:
          cp -v bin/libQupZilla*.dylib tests/autotests;
     fi
   - cd scripts && ./run_tests.sh $QMAKE
+  - ccache --show-stats
 
 before_deploy:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then


Martii:
@Chocobo1 
Would this help speed things up on the Mac directly? e.g. the `brew install ccache;` and `PATH` modifications?

muesli:
How is this supposed to speed up anything? The cache would be empty everytime a new Travis build starts.

On the other hand: would you even want it to cache anything? This is a CI, its results should be deterministic, and ccache makes it less so.

Martii:
@muesli
https://travis-ci.org/QupZilla/qupzilla/builds is what I'm watching to see if it's helpful *(time spent)*. Some days I dread building on my remote Mac and any speed improvements would be welcome. I did just figure out that using "Remote Management" with VNC enable is faster than "Screen Sharing"... but again **any** improvement would be extremely welcome.

muesli:
@Martii 

I guess my point is: how is this supposed to speed up anything? This would work on your own build machine, yes, but I can't see how it would cache anything on Travis (see 5% cache-hit rate).

Chocobo1:
>Would this help speed things up on the Mac directly? 

Of course,
This is before the PR: https://travis-ci.org/QupZilla/qupzilla/jobs/212479759#L910
Time spent in make: 691sec

This is after the cache is fully populated: https://travis-ci.org/Chocobo1/qupzilla/jobs/212635156#L936
Time spent in make: 87sec

>The cache would be empty everytime a new Travis build starts.

https://docs.travis-ci.com/user/caching#How-does-caching-work%3F

>On the other hand: would you even want it to cache anything? 

Because you will find build failure faster and save time waiting for it.
And possibly save some resource which is shared by all CI users.

>This is a CI, its results should be deterministic, and ccache makes it less so.

Correct, can't argue with this.
But I've never meet a build failure that I need to clear the cache to fix it.

muesli:
@Chocobo1 Thanks for clarifying!

Martii:
@Chocobo1 
Yes thank you very much. I'll go install that now for direct builds.

Martii:
@Chocobo1 
One other quickie... I see the spec on `make` says:

``` sh-session
If the -j option is given without an argument, make will not limit the number of jobs that can run simultaneously.
```

.... would this be a future proofing in case Travis CI here bumps up the maximum allowed? e.g. would that work okay?

muesli:
@Martii This (for most builds) will drag the machine to a complete stand-still. I'd recommend setting it to the number of available CPU cores (potentially *2 for HT).

Martii:
@muesli 
I see that now... my VNC is somewhat frozen. ;) Just thought I'd get the official word. :smile_cat:

Chocobo1:
>e.g. would that work okay?

In short, no.
I remembered there was a guy with a building farm ~60 machines or cores, can't remember, he faced some sort of racing condition when building qbittorrent with `make -j`.

muesli:
@Martii Heh, it can work fine for small C builds. For Qt apps, though, it will result in spawning a gazillion moc processes parallely. That usually kills the machine.

Martii:
Cool. Thanks. Looks like I don't need the `PATH` mods here. The cellar for the `.../bin` folder is somehow already in the path. e.g. can run `ccache` with no changes to `.bash_profile`. I'll twiddle with different values with `make` to see what I can balance between my remote session and the capabilities of the machine. 

Thank you both for your advice.

Martii:
For future direct implementation reference with static paths:

```
$ brew install ccache
```

~/.bash_profile
``` bash
export CCACHE="/usr/local/Cellar/ccache/3.3.4.1"
export CC="ccache gcc"
export CXX="ccache g++"

export QTDIR="$HOME/Qt/5.8/clang_64"

export PATH="$CCACHE/libexec:$QTDIR/bin:$PATH"
```

and:

``` sh-session
$ which gcc
/usr/local/Cellar/ccache/3.3.4.1/libexec/gcc

$ which g++
/usr/local/Cellar/ccache/3.3.4.1/libexec/g++

$ qmake
$ make -j2 CC="$CC" CXX="$CXX"
```

OS X 10.12.3 *(Sierra)*

PR #2264, opened 2017-03-20
Unix: Allow git hooks to work for development on all related nixes including Mac

Martii:
* Same feature already optional in Linux via https://github.com/QupZilla/qupzilla/blob/a71302fbc6f91aaf75ac8afca7f62227ea5ef767/git_hooks/hooks-install.sh
* Windows should probably have this too for developers but don't have MSVS installed to test and won't for quite some time.

Post followup for #1958

---
**NOTE**: PR'd to `master` so I presume it won't show up until 2.2.0

About with hash

@srazi
Please take a look at this image and look at the speed dial on the Mac. It's been this way for a really long time although the broken-image image changed. I have nothing special proxy wise e.g. direct TCP/IP and clean profile. Linkage works but just no images on the Mac like there is Linux and Windows. Thanks.

### DIFF
--- a/src/defines.pri
+++ b/src/defines.pri
@@ -47,13 +47,7 @@ win32-msvc* {
 }
 
 # QtDBus not available on Mac
-mac {
-    DEFINES *= DISABLE_DBUS
-
-    # Git revision
-    rev = $$system(cd ../ && sh $$PWD/../scripts/getrevision.sh)
-    !equals(rev, ""): DEFINES *= GIT_REVISION=\\\"""$$rev"\\\""
-}
+mac: DEFINES *= DISABLE_DBUS
 
 haiku-* {
     DEFINES *= DISABLE_DBUS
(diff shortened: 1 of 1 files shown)


Martii:
PR READY unless suggestions roll by

src/defines.pri
```diff
@@ -47,7 +47,13 @@ win32-msvc* {
 }
 
 # QtDBus not available on Mac
-mac: DEFINES *= DISABLE_DBUS
+mac {
+    DEFINES *= DISABLE_DBUS
+
+    # Git revision
```
nowrep: Instead of copying the same code from unix conditional, move it to conditional that is for both unix & mac.

srazi:
> Please take a look at this image and look at the speed dial on the Mac. It's been this way for a really long time although the broken-image image changed...

Please create a new issue for this one, After some testing I found that `qml/QtWebEngine` and `qml/QtQuick` should added to bundle. but they have sub-folders and each sub-folder has a `dylib` that `install_name_tool` should applied on them. :|

Martii:
https://travis-ci.org/QupZilla/qupzilla/builds/213117653

>  Ran for 32 min 51 sec
Total time 24 min 41 sec 

That doesn't make sense. :\

Martii:
@muesli 
Thanks for that clarification.

PR #2273, opened 2017-03-21
Open URLs from "Go to web address" (contextmenu) in a new tab

ThaFireDragonOfDeath:
Problem:
The actions from the context menu: "Translate page", "Validate page", "Search ... with Google" and "Show source code" opens the action page in a new selected tab. "Go to web address" instead, opens the selected URL in the same tab. For the users, that want to read the rest of the origin page is it better, when the new page opens in a new tab.

Changes:
The "Go to web address" action opens the selected text link in a new selected tab.

### DIFF
--- a/src/lib/webengine/webview.cpp
+++ b/src/lib/webengine/webview.cpp
@@ -522,7 +522,7 @@ void WebView::openUrlInNewTab(const QUrl &url, Qz::NewTabPositionFlags position)
 void WebView::openActionUrl()
 {
     if (QAction* action = qobject_cast<QAction*>(sender())) {
-        load(action->data().toUrl());
+        openUrlInNewTab(action->data().toUrl(), Qz::NT_SelectedTab);
     }
 }
 


src/lib/webengine/webview.cpp
```diff
@@ -522,7 +522,7 @@ void WebView::openUrlInNewTab(const QUrl &url, Qz::NewTabPositionFlags position)
 void WebView::openActionUrl()
 {
     if (QAction* action = qobject_cast<QAction*>(sender())) {
-        load(action->data().toUrl());
+        openUrlInNewTab(action->data().toUrl(), Qz::NT_SelectedTab);
```
nowrep: This has nothing to do with "go to web address" action, you just changed the behavior of `openActionUrl` method that is used for more action that just go to web address.

As you can see on line 871, you can middle-click the action to open in new tab.

ThaFireDragonOfDeath: Oh, with the the possibility to open the URL in a new tab with a middle click, this change makes no sense. I think this can be closed (or is there anything else to do in this context?)...

PR #2274, opened 2017-03-23
Repair Mac Like Theme icon so it's not clipped

Martii:


Martii:
PR READY

Martii:
Tried a few dozen ways of improving this and finally ended up flipping it mid-axis. Mac displays these as 16x16.

Anyhow... should be 32x32 now... PR READY. Sorry for the extra commit.

PR #2280, opened 2017-03-26
Refactor some codes related to sessions.

srazi:


src/lib/app/mainapplication.cpp
```diff
@@ -403,17 +403,17 @@ MainApplication::AfterLaunch MainApplication::afterLaunch() const
...
     setOverrideCursor(Qt::BusyCursor);
 
+    if (!window)
+        window = createWindow(Qz::BW_OtherRestoredWindow);
+
     window->setUpdatesEnabled(false);
-    window->tabWidget()->closeRecoveryTab();
+
+    if (m_isRestoring)
+        window->tabWidget()->closeRecoveryTab();
```
nowrep: Shouldn't this be called everytime? If there is no recovery tab it does nothing anyway.

srazi: Well with current implementation of `session manager` if user open another session other than the last session s/he don't want that the recovery page closed automatically.

PR #2292, opened 2017-04-04
Fix 2178

srazi:


iondissonance:
Thanky you very much!

src/lib/other/checkboxdialog.cpp
```diff
@@ -16,47 +17,28 @@
...
-    ui->iconLabel->setPixmap(icon.pixmap(48, 48));
-    ui->iconLabel->setFixedWidth(48);
-}
-
-void CheckBoxDialog::setText(const QString &text)
-{
-    ui->textLabel->setText(text);
+    setCheckBox(m_checkBox);
+    setStandardButtons(buttons);
+    setWindowFlags(windowFlags() | Qt::MSWindowsFixedSizeDialogHint | Qt::WindowTitleHint | Qt::WindowSystemMenuHint | Qt::WindowCloseButtonHint);
```
nowrep: Can you make this only for Q_OS_WIN? I think it may not work that well on other platforms (mainly on Wayland).

srazi: Did you mean inheriting from QMessageBox or `setWindowFlags()`? As `setWindowFlags()` uses what was used in old `CheckBoxDialog` code.

nowrep: Ah, true, sorry.

PR #2293, opened 2017-04-04
Session manager

srazi:
***This is Session Manager as a core feature.***
1. The `$CurrentProfile/session.dat` is used as default session and a fallback for non-existing session files (on startup).
2. The `$CurrentProfile/session.dat.old` and `$CurrentProfile/session.dat.old1` are same as before i.e. backup from last active session and backup from last backup.
3. All other session files are stored in `$CurrentProfile/sessions`
4. `Open` action will open windows/tabs in current active session.
5. `Switch To` action will save current session -> close all windows ->then load selected session
6. User can not switch to backup sessions as they are overwritten by application.
7. Also deleting of backup sessions and default session is not available.
8. `Select session` is added to `after launch` that will ask about startup session after starting QupZilla.

**P.S.** I don't like the last commit  338e552 but I could not find a better solution.

### DIFF
--- a/src/lib/session/sessionmanager.cpp
+++ b/src/lib/session/sessionmanager.cpp
@@ -23,12 +23,16 @@
 #include "settings.h"
 
 #include <QAction>
+#include <QComboBox>
 #include <QDateTime>
+#include <QDialogButtonBox>
 #include <QDir>
 #include <QFileSystemWatcher>
 #include <QInputDialog>
+#include <QLabel>
 #include <QMenu>
 #include <QMessageBox>
+#include <QVBoxLayout>
 
 
 SessionManager::SessionManager(QObject* parent)
@@ -385,4 +389,42 @@ void SessionManager::autoSaveLastSession()
 
     saveSettings();
     writeCurrentSession(m_lastActiveSessionPath);
+}
+
+QString SessionManager::askSessionFromUser()
+{
+    fillSessionsMetaDataListIfNeeded();
+
+    QDialog dialog(mApp->getWindow(), Qt::WindowStaysOnTopHint);
+    QLabel label(tr("Please select the startup session:"), &dialog);
+    QComboBox comboBox(&dialog);
+    QDialogButtonBox buttonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel, &dialog);
+    connect(&buttonBox, &QDialogButtonBox::accepted, &dialog, &QDialog::accept);
+    connect(&buttonBox, &QDialogButtonBox::rejected, &dialog, &QDialog::reject);
+
+    QVBoxLayout layout;
+    layout.addWidget(&label);
+    layout.addWidget(&comboBox);
+    layout.addWidget(&buttonBox);
+    dialog.setLayout(&layout);
+
+
+    const QFileInfo lastActiveSessionFileInfo(m_lastActiveSessionPath);
+
+    for (const SessionMetaData &metaData : m_sessionsMetaDataList) {
+        if (QFileInfo(metaData.filePath) != lastActiveSessionFileInfo) {
+            comboBox.addItem(metaData.name, metaData.filePath);
+        }
+        else {
+            comboBox.insertItem(0, tr("%1 (last session)").arg(metaData.name), metaData.filePath);
+        }
+    }
+
+    comboBox.setCurrentIndex(0);
+
+    if (dialog.exec() == QDialog::Accepted) {
+        m_lastActiveSessionPath = comboBox.currentData().toString();
+    }
+
+    return m_lastActiveSessionPath;
 }
--- a/src/lib/session/sessionmanager.cpp
+++ b/src/lib/session/sessionmanager.cpp
@@ -260,6 +260,34 @@ void SessionManager::saveSession()
     }
 }
 
+void SessionManager::newSession()
+{
+    bool ok;
+    QString sessionName = QInputDialog::getText(mApp->activeWindow(), tr("New Session"),
+                                         tr("Please enter a name to create new session:"), QLineEdit::Normal,
+                                         tr("New Session (%1)").arg(QDateTime::currentDateTime().toString("dd MMM yyyy HH-mm-ss")), &ok);
+
+    if (ok) {
+        const QString filePath = QString("%1/%2.dat").arg(DataPaths::path(DataPaths::Sessions)).arg(sessionName);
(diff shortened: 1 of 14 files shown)


src/lib/session/sessionmanager.cpp
```diff
@@ -0,0 +1,430 @@
...
+        sessionFilePath = action->data().toString();
+    }
+
+    bool ok;
+    const QString suggestedName = QFileInfo(sessionFilePath).baseName() + (clone ? tr("_cloned") : tr("_renamed"));
+    QString newName = QInputDialog::getText(mApp->activeWindow(), (clone ? tr("Clone Session") : tr("Rename Session")),
+                                            tr("Please enter a new name:"), QLineEdit::Normal,
+                                            suggestedName, &ok);
+
+    if (ok) {
```
nowrep: early return here `if (!ok) return`

src/lib/session/sessionmanager.cpp
```diff
@@ -0,0 +1,430 @@
...
+            return;
+        }
+
+        if (clone) {
+            if (!QFile::copy(sessionFilePath, newSessionPath)) {
+                QMessageBox::information(mApp->activeWindow(), tr("Error!"), tr("An error occurred when cloning session file."));
+                return;
+            }
+        }
+        else {
```
nowrep: else if

src/lib/session/sessionmanager.cpp
```diff
@@ -0,0 +1,430 @@
...
+}
+
+void SessionManager::saveSession()
+{
+    bool ok;
+    QString sessionName = QInputDialog::getText(mApp->activeWindow(), tr("Save Session"),
+                                         tr("Please enter a name to save session:"), QLineEdit::Normal,
+                                         tr("Saved Session (%1)").arg(QDateTime::currentDateTime().toString("dd MMM yyyy HH-mm-ss")), &ok);
+
+    if (ok) {
```
nowrep: early return

src/lib/session/sessionmanager.cpp
```diff
@@ -0,0 +1,430 @@
...
+}
+
+void SessionManager::newSession()
+{
+    bool ok;
+    QString sessionName = QInputDialog::getText(mApp->activeWindow(), tr("New Session"),
+                                         tr("Please enter a name to create new session:"), QLineEdit::Normal,
+                                         tr("New Session (%1)").arg(QDateTime::currentDateTime().toString("dd MMM yyyy HH-mm-ss")), &ok);
+
+    if (ok) {
```
nowrep: early return

src/lib/session/sessionmanager.cpp
```diff
@@ -0,0 +1,430 @@
...
+        }
+
+        m_lastActiveSessionPath = filePath;
+        autoSaveLastSession();
+    }
+}
+
+void SessionManager::fillSessionsMetaDataListIfNeeded()
+{
+    if (m_sessionsMetaDataList.isEmpty()) {
```
nowrep: early return

src/lib/tabwidget/tabwidget.cpp
```diff
@@ -787,8 +787,17 @@ QByteArray TabWidget::saveState()
         if (!webTab)
             continue;
 
-        WebTab::SavedTab tab(webTab);
-        tabList.append(tab);
+        // save state does not support saving restore tab
```
nowrep: What does this mean?

src/lib/session/sessionmanager.cpp
```diff
@@ -0,0 +1,430 @@
...
+    connect(&buttonBox, &QDialogButtonBox::accepted, &dialog, &QDialog::accept);
+    connect(&buttonBox, &QDialogButtonBox::rejected, &dialog, &QDialog::reject);
+
+    QVBoxLayout layout;
+    layout.addWidget(&label);
+    layout.addWidget(&comboBox);
+    layout.addWidget(&buttonBox);
+    dialog.setLayout(&layout);
+
+
```
nowrep: extra newline

src/lib/tabwidget/tabwidget.cpp
srazi: I mean save state just saves the url of restore tab (i.e. qupzilla:restore) and not the session data loaded in restore tab or at least the path to loaded session file.

This workaround fixes saving sessions when a restore tab is opened. Another solution is closing restore tab but it can change current index and also sometimes closing restore tab will lead to a window without tab.

nowrep: Yeah, I don't really like it.
Please remove this check for this patch and let's implement it correctly in followup commit - eg. skip the restore tab (or any other specified URLs) when saving the session properly.

srazi: OK. How about this fix: (maybe when `tabList.isEmpty()` we should append a new tab?)
```diff
diff --git a/src/lib/tabwidget/tabwidget.cpp b/src/lib/tabwidget/tabwidget.cpp
index 18e4858..7975f03 100644
--- a/src/lib/tabwidget/tabwidget.cpp
+++ b/src/lib/tabwidget/tabwidget.cpp
@@ -782,16 +782,28 @@ QByteArray TabWidget::saveState()
 {
     QVector<WebTab::SavedTab> tabList;
 
+    int currentTabIndex = currentIndex();
+
     for (int i = 0; i < count(); ++i) {
         WebTab* webTab = weTab(i);
         if (!webTab)
             continue;
 
+        if (webTab->url().toString() == QLatin1String("qupzilla:restore")) {
+            if (i < currentTabIndex)
+                --currentTabIndex;
+            continue;
+        }
+
         WebTab::SavedTab tab(webTab);
         tabList.append(tab);
     }
 
     QByteArray data;
+
+    if (tabList.isEmpty())
+        return data;
+
     QDataStream stream(&data, QIODevice::WriteOnly);
 
     stream << tabList.count();
@@ -800,7 +812,7 @@ QByteArray TabWidget::saveState()
         stream << tab;
     }
 
-    stream << currentIndex();
+    stream << currentTabIndex;
 
     return data;
 }
```

nowrep: Something along those lines, but not quite this :D

srazi: @nowrep So I hope you will fix it :D , maybe we have to skip `view-source` tabs, too. As an example QupZilla crashes when switching to a restored view-source tab of a youtube video.

Issue #2139, opened 2016-12-10
Tab Manager. More extensible.


DebugReport:
Maybe add switching with single click on tab?
And change switching with Ctrl+Tab/Ctrl+Shift+Tab as visit?
If open github tab, and then switch to other tab, press on Ctrl+Shift+Tab return to github tab.
And maybe highlighting recently openning tabs?

robberer:
> Maybe add switching with single click on tab?

I would like to see this implemented, too

DebugReport:
And what about moving tabs in Tab Manager? I think it's good idea... :-)

robberer:
I want to use the TabManager as replacement for the main TabBar to get more height on my widescreen, and because usually I have more than 30 tabs open which become unreadable in the main tab bar.

So I'm more interested in switching tabs in the manager with a single click.

But if I would want to use both (tab manager + tab bar) then moving tabs could be useful ;-)

PR #2297, opened 2017-04-07
Improve TabManager

srazi:
***Some improvement so TabManager can be a better replacement for main tabbar.***

1. Single click to switch between tabs #2139
2. Add `close` and `add` tab button to easily close or add tabs.

tabmanager


nowrep:
Looks fine.

On an unrelated note, isn't there a way to speedup the Mac builds? Downloading 1GB installer in each build is *really* bad.

srazi:
> On an unrelated note, isn't there a way to speedup the Mac builds? Downloading 1GB installer in each build is really bad.

Well, indeed my next PR is about `CI`s will send it, ASAP.

Martii:
> isn't there a way to speedup the Mac builds? Downloading 1GB installer in each build is really bad.

It's probably because the unified installer does the debug libs as well. So using `brew` will probably only allow release testing in CI....and of course the entire dependencies will have to be re `install_name_tool`ed. If you can figure out a release target to original Qt with no debug libs that could work although will limit Travis CI.

PR #2301, opened 2017-04-07
Ci faster mac win builds

srazi:
**OSX:** As we just deploy on pushing new tags so all other builds can use Homebrew qt that is a much smaller file to download.
**Windows:** As we just deploy on pushing new tags so for all other builds we don't need to make installer.

**‌Before:**

win_before

osx_before

**‌After:**

win_after

osx_after


nowrep:
Thanks!

Martii:
Kewl switch for Travis CI :)

PR #2330, opened 2017-05-08
Update web page zoom factor from preferences

bvbfan:
Workaround Qt5.8 zoom factor bug

Fix bug: 2289

Signed-off-by: Anthony Fieroni <bvbfan@abv.bg>

### DIFF
--- a/src/lib/webengine/webview.cpp
+++ b/src/lib/webengine/webview.cpp
@@ -40,6 +40,7 @@
 
 #include <QDir>
 #include <QTimer>
+#include <QVersionNumber>
 #include <QDesktopServices>
 #include <QWebEngineHistory>
 #include <QClipboard>
@@ -153,6 +154,15 @@ void WebView::setPage(WebPage *page)
 
     // Set default zoom level
     zoomReset();
+
+    static const bool zoomBug = QVersionNumber::fromString(qVersion()) < QVersionNumber(5, 9, 0);
+    if (zoomBug) {
+        connect(m_page, &WebPage::loadProgress, this, [this](int progress) {
+            if (progress > 0) {
+                applyZoom();
+            }
+        });
+    }
 
     // Actions needs to be initialized for every QWebEnginePage change
     initializeActions();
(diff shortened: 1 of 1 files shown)


src/lib/tabwidget/tabwidget.cpp
```diff
@@ -194,6 +194,15 @@ void TabWidget::loadSettings()
     m_newEmptyTabAfterActive = settings.value("newEmptyTabAfterActive", false).toBool();
     settings.endGroup();
 
+    settings.beginGroup("Web-Browser-Settings");
```
nowrep: I don't think this is wanted, it will overwrite any previously set zoom when applying preferences.

src/lib/webengine/webview.cpp
```diff
@@ -154,6 +159,15 @@ void WebView::setPage(WebPage *page)
     // Set default zoom level
     zoomReset();
 
+    // Workaround < Qt5.9: apply zoom to every page loading progress
+    if (QVersionNumber::fromString(qVersion()) < QVersionNumber(5, 9, 0)) {
```
nowrep: Just use `#if` or cache the result in static variable.

src/lib/tabwidget/tabwidget.cpp
bvbfan: I think it's expected behavior when it change default zoom factor to reflect in all open tabs, no?

nowrep: I wouldn't expect it to behave like that, it's the "Default" zoom for new tabs.

src/lib/webengine/webview.cpp
bvbfan: I'm not prefer preprocessor directive cause Qupzilla can be compiled against different Qt version than runtime, qVersion() should return runtime version. I'll do it in static variable.

bvbfan: Did you test Qt5.9 it's apply correct zoom factor?

src/lib/webengine/webview.cpp
```diff
@@ -145,6 +146,10 @@ void WebView::setPage(WebPage *page)
         return;
     }
 
+    if (m_page) {
+        m_page->deleteLater(); // we are the parent, see below
```
bvbfan: ok

PR #2338, opened 2017-05-24
Add „configure like“ source code compilation script.

cranes-bill:


nowrep:
Can you please rename it to build-appimage.sh, or even `Recipe` as is used for AppImage scripts?

PR #2347, opened 2017-06-07
Ensure transient parent window for LocationCompleterView

### LINKED ISSUE #2344: [Wayland] URL bar dropdown menu misplaced
To reproduce in a Plasma Wayland session:
1. Focus the URL lineedit
2. Type something
3. The dropdown menu will open in a random position, rather than below the line edit.

elvisangelaccio:
This makes sure that the completer listview is properly positioned on
wayland. Tested with Qt 5.9 RC.

Closes #2344.

### DIFF
--- a/src/lib/navigation/completer/locationcompleter.cpp
+++ b/src/lib/navigation/completer/locationcompleter.cpp
@@ -29,6 +29,8 @@
 #include "bookmarkitem.h"
 #include "qzsettings.h"
 
+#include <QWindow>
+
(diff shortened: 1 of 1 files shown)


elvisangelaccio:
hmm @nowrep please ignore this PR for now, there is other stuff broken in my wayland session which used to work before. So I might have something broken on my end.

elvisangelaccio:
Ok, after some more testing I can confirm that this patch works for me.

PR #2393, opened 2017-07-26
Add missing RUNPATH so that qupzilla finds its shared lib

dfaure:
when installed into a custom prefix, on Linux.

If qupzilla used cmake and ECM, this would have worked out of the box :-)

I actually volunteer to port qupzilla to cmake (qmake + env vars is rather cumbersome), but I would only test it on Linux, it would have to be tested by others on other platforms.

(Alternatively if qmake is kept, we could port to the Qt 5.8 configure system to improve upon the env var solution.)

### DIFF
--- a/src/main/main.pro
+++ b/src/main/main.pro
@@ -40,4 +40,10 @@ openbsd-*|freebsd-*|haiku-* {
 
 include(../install.pri)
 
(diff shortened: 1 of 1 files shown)


nowrep:
> I actually volunteer to port qupzilla to cmake (qmake + env vars is rather cumbersome), but I would only test it on Linux, it would have to be tested by others on other platforms.

That would be great, I don't like it either (but don't hate it enough to port to cmake myself :)
If ECM works on Windows and Mac we can use that too.

PR #2426, opened 2017-08-22
Windows: At first try to show native default app settings UI (>=Vista).

srazi:
The main part of this patch has been taken from Firefox source code ([here](https://github.com/mozilla/gecko-dev/blob/master/browser/components/shell/nsWindowsShellService.cpp#L364)) that is licensed under MPL-2.0. Is it OK merging MPL code into GPL code?

I tested it under Windows 10 and WIndows server 2012 (that has same API like win8).

### DIFF
--- a/src/lib/other/registerqappassociation.cpp
+++ b/src/lib/other/registerqappassociation.cpp
@@ -1,5 +1,5 @@
 /* ============================================================
-* Copyright (C) 2012-2014  S. Razi Alavizadeh <s.r.alavizadeh@gmail.com>
+* Copyright (C) 2012-2017  S. Razi Alavizadeh <s.r.alavizadeh@gmail.com>
 * This file is part of QupZilla - WebKit based browser 2010-2014
 * by  David Rosca <nowrep@gmail.com>
 *
@@ -148,6 +148,12 @@ bool RegisterQAppAssociation::isVistaOrNewer()
             QSysInfo::windowsVersion() <= QSysInfo::WV_NT_based);
 }
 
+bool RegisterQAppAssociation::isWin10OrNewer()
+{
+    return (QSysInfo::windowsVersion() >= QSysInfo::WV_WINDOWS10 &&
+            QSysInfo::windowsVersion() <= QSysInfo::WV_NT_based);
+}
+
 void RegisterQAppAssociation::registerAssociation(const QString &assocName, AssociationType type)
 {
     if (isVistaOrNewer()) { // Vista and newer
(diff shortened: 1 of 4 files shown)


src/lib/other/registerqappassociation.cpp
```diff
@@ -261,6 +267,65 @@ void RegisterQAppAssociation::registerAllAssociation()
...
+                                      IID_IApplicationActivationManager,
+                                      (void**)&pActivator);
+
+        if (!SUCCEEDED(hr)) {
+            return false;
+        }
+
+        DWORD pid;
+        hr = pActivator->ActivateApplication(
+            L"windows.immersivecontrolpanel_cw5n1h2txyewy"
```
nowrep: What does this `cw5n1h2txyewy` mean?

nowrep:
> Is it OK merging MPL code into GPL code?

I'm not really sure if this even needs to be copyrighted, it is just using few WinAPI functions.

src/lib/other/registerqappassociation.cpp
srazi: The method [IApplicationActivationManager::ActivateApplication()](https://msdn.microsoft.com/en-us/library/windows/desktop/hh706903(v=vs.85).aspx) activates a specified Windows Store app  by its appUserModelId.
The appUserModelId of `Settings` is:
`windows.immersivecontrolpanel_cw5n1h2txyewy!microsoft.windows.immersivecontrolpanel`

Some other appUserModelIds are as follows:
```cpp
windows.immersivecontrolpanel_cw5n1h2txyewy!microsoft.windows.immersivecontrolpanel
Microsoft.MicrosoftEdge_8wekyb3d8bbwe!MicrosoftEdge
Microsoft.MicrosoftEdge_8wekyb3d8bbwe!ContentProcess
Microsoft.MicrosoftEdge_8wekyb3d8bbwe!BCHost
Microsoft.Windows.ContentDeliveryManager_cw5n1h2txyewy!App
Microsoft.Windows.Photos_8wekyb3d8bbwe!App
```
As you see `cw5n1h2txyewy` and `8wekyb3d8bbwe` have been repeated. So I think they are somethings like a product-category id for a company registered in Windows store.

nowrep: Alright, thanks for the explanation.

srazi:
> I'm not really sure if this even needs to be copyrighted, it is just few WinAPI calls. Most likely a similar snippet can be found on msdn.

Yeah, I agree. Indeed, after some more reading I found out that it is just how one supposed to launch a Windows store app and pass some arguments to it.

So, I will close this PR, and after removing license notice, will commit it.

nowrep:
Too late, but do it in followup commit :)
Also, you can just update the pull request with git push -f

PR #2461, opened 2017-09-21
get more native look

Jacobtey:


nowrep:
I can agree only with "network-connect" and "network-disconnect", other icon names means completely different thing and it will only look good in your theme.

Jacobtey:
> I can agree only with "network-connect" and "network-disconnect", other icon names means completely different thing and it will only look good in your theme.

Hm. It's up to you. But otherwise it won't look good anywhere. StatusBarIcons can't adopt custom themes.

PR #2468, opened 2017-10-06
Mac: Revert #2184 for Qt 5.9.1+

Martii:
* Now that upstream Qt has support for this back out the changes.

NOTE(S)
* Please note that users with Qt 5.9.0 and earlier on MacOS the bug will reappear... so minimum Qt build should be 5.9.1 if those icons in the dropdown menus are wanted

Ref(s):
* https://github.com/QupZilla/qupzilla/pull/2184#issuecomment-308790083

### DIFF
--- a/mac/macdeploy.sh
+++ b/mac/macdeploy.sh
@@ -13,7 +13,6 @@ MACDEPLOYQT=$1
 QTDIR="`dirname $MACDEPLOYQT`/.."
 LIBRARY_NAME="libQupZilla.2.dylib"
 PLUGINS="QupZilla.app/Contents/Resources/plugins"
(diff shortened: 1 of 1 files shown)


Martii:
@nowrep 
I am signed up on KDE's identity site but have no clue how to fork yet or even submit a PR.

Any good, recommended, primers for that site? 

Also do you need to add me to something *(the project?)*. 

Btw my SourceForge email address is in effect now and the old email address is no longer owned by myself.

nowrep:
https://community.kde.org/Infrastructure/Phabricator

You don't fork repo as with github, instead you submit patches.

PR #2523, opened 2017-12-08
Mac: Repair missing hidden `.VolumeIcon.icns`

Martii:
* `osascript` deletes the icns file after completion for no apparent reason in High Sierra
* Move `iconutil` after this event

NOTE:
* This shouldn't cause a regression in older OS X's


Martii:
GH isn't allowing me to merge as a bug fix into 2.2.2... will make a new one to master then.

PR #2539, opened 2018-01-08
 Style Update [SpeedDial+SpeedDial:Icons]

Diablopc:


nowrep:
I like it, but there is an issue with loading gif not being properly resized for very small dials (eg. set size to minimum = 100).

Diablopc:
>  loading gif not being properly resized for very small dials

Hah, looks realy funny o_0

Yep, thats my bug. I haven't check this moments. 
Agree this is realy need some rework. I'll do it tomorrow...

cranes-bill:
@nowrep Now is the moment to consider merging `start` page with `speeddial`?

Diablopc:
@cranes-bill it's planned, but not all at once =)

Diablopc:
> loading gif not being properly resized 

Fixed by 5d76be6

> It's fine those buttons being big for very small dials, but they should also resize when on very big dials (eg. maximum size = 500).

Fixed by a60dd0c

- Min button size set to 16. 
- Max button size not limited and depends of dial size


Also:
- Added default background
- Fixed incorrect tiles enumeration [[7bc8123]]
- Fully switched to scalable graphics

@nowrep what U think about?

nowrep:
Perfect

Diablopc:
@nowrep 
Pls, update this file in yours repo.
https://github.com/Diablopc/qupzilla/blob/master/src/lib/data/html/sd_bg.svg
It was broken =(

cranes-bill:
Why the modal dialog overfills the window!?
capture du 2018-01-11 08 35 43
And if i enable the bookmarks bar, it's getting badly:
capture du 2018-01-11 08 37 40

Diablopc:
@cranes-bill already working on this!

nowrep:
This is a long-standing issue, maybe we can move background selection to second column?

Diablopc:
@nowrep I've already added a scrollbar for cases where content can't fit to screen.
Commit will be in one hour

nowrep:
I'd rather have it in two columns instead of adding scrollbars.

Diablopc:
@nowrep But then we must add a horisontal scroll due to portrait screens...

I have some idea how to decrease vertical size without layout changes...
Wait aproximately a hour, i'll try to do that

Diablopc:
@nowrep 
Here is it


Scrollbar appears only if active area lower than 550px.
This is enough to not show it on low resolution screens inclusive 768px height

nowrep:
Ok, that looks fine. Can you ale please hide size of pages slider when "change size of pages" is unchecked? It's kind of confusing as the disabled state is not easy to distinguish.

Diablopc:
> Can you also please hide size of pages slider when "change size of pages" is unchecked?

Sure, it planned.
In the nearest future planned to move "Add" button to dial space and restile sliders to common style, also with this i'll hide unused messages =)

nowrep:
I need your full name, otherwise I can't push this to Falkon repo. And that applies for all following commits too, please commit with full name.

Diablopc:
Ou, that's not a problem =)
I hope it enough just to add this to my profile?

nowrep:
Alright, thanks.

>  I hope it enough just to add this to my profile?

Only if you commit through github webpage. You need to set it in your git config.

Diablopc:
> You need to set it in your git config.

Ok. It  will be done before next commit.

cranes-bill:
Noticed onother one unintended behaviour:
I have `Don't close window upon closing last tab` option  set. So when i close the last one tab, the result is an empty Speed dial.
capture du 2018-01-12 22 12 07

Diablopc:
> @cranes-bill I have Don't close window

it looks weird .... I'm use this option to, but never noticed such behavior...
btw try to pick it up from my repo, there was a lot of changes that not applied in main repo

nowrep:
@cranes-bill That has nothing to do with this, you should open separate issue for this.

cranes-bill:
> ..., you should open separate issue for this.

That was my initial plan, but due to i'm avoiding of „non sense issues“, decided posting here.
But now, since you tel, i will open a new issue.

PR #2556, opened 2018-01-18
SpeedDial: Add searchbar

Diablopc:
- Disabled

- Enabled

- In config


Martii:
@Diablopc 
Best to make a branch on your repository, and ensure that your master is up to date before that so comments don't reappear that aren't applicable here. e.g. mine. Thanks.

cranes-bill:
Fine, but i think it will look better with square borders (just like the other stuff here).

nowrep:
Alright, as @Martii said, please make this pull request again from clean branch (it's also good idea to use feature branches for all future pull requests).

Also, instead of configurable engine, let's just make it use default search engine.

Diablopc:
@cranes-bill 
Like this?

> @nowrep pull request again from clean branch

Ok

> let's just make it use default search engine.

Hmm, idk how much it will be convenient because default search engine available from address bar.

cranes-bill:
Yes square**d** :blush:, but that is only in my opinion.

PR #2558, opened 2018-01-18
SpeedDial: search

ghost:

### DIFF
--- a/src/lib/data/html/speeddial.html
+++ b/src/lib/data/html/speeddial.html
@@ -321,14 +321,6 @@ div.entry:hover .edit, div.entry:hover .close, div.entry:hover .reload {
     width: 25px;
 }
 
-.sesel {
-    margin: 2px;
-    padding: 3px 0;
-    height: 45px;
-    vertical-align: middle;
-    border-bottom: 1px solid rgba(135,135,135, 1.0);
-}
-
 #ImgSelector {
     position:relative;
     z-index: 1;
@@ -561,10 +553,10 @@ input[type=checkbox] {
     outline: none;
     background: rgba(65,65,65, 1.0);
     font-weight: bolder;
-    color: rgba(236,236,236, 1.0);
+    color: rgba(118,118,118, 1.0);
 }
 
-.submit {
+#submit {
     width: 30px;
     height: 30px;
     background: url("qrc:html/search.svg") no-repeat center center;
@@ -573,9 +565,12 @@ input[type=checkbox] {
     border: 0;
     outline: none;
     float: right;
+    margin: 0;
+    outline: none;
+    box-shadow: none;
 }
 
-.submit:hover {
+#submit:hover {
     background: url("qrc:html/search_active.svg") no-repeat center center;
 }
 
@@ -609,17 +604,12 @@ $(document).keyup(function(e) {
 });
 
 function searchToggle() {
-    if ("%SEARCH-ENABLED%" == '1') {
-        external.speedDial.setsearchEnabled(false);
+    if ("%SEARCH-ENABLED%" == 'true') {
+        external.speedDial.setSearchEnabled(false);
     } else {
-        external.speedDial.setsearchEnabled(true);
+        external.speedDial.setSearchEnabled(true);
     };
     location.reload();
-}
-
-function setEngine() {
-    engine = $('#SearchEngineSel').val()
-    external.speedDial.setsearchEngine(engine);
 }
 
 function emitChanged(pages) {
@@ -645,7 +635,6 @@ function configureSpeedDial() {
     $('#BgImgToggle').is(':checked') ? $('#BgImgSel').removeAttr('disabled') : $('#BgImgSel').attr('disabled', 'disabled');
     $('#BgImgToggle').is(':checked') ? $('#BgImgSelSiz').removeAttr('disabled') : $('#BgImgSelSiz').attr('disabled', 'disabled');
     $('#SdSizeToggle').is(':checked') ? $('#SdSize').removeAttr('disabled') : $('#SdSize').attr('disabled', 'disabled');
-    $('#SearchEngineSel').val('%SEARCH-ENGINE%').attr('selected', 'selected');
     if ($('#BgImgToggle').prop('checked') != true) {
         $('#ImgSelectorMenu').css({'color' : 'rgba(0,0,0, 0.0)', 'text-shadow' : 'none'});
         $('#BgImgSel').css({'color' : 'rgba(0,0,0, 0.0)', 'text-shadow' : 'none'});
@@ -934,7 +923,6 @@ function saveSettings() {
     external.speedDial.setPagesInRow(MAX_PAGES_ROW);
     external.speedDial.setSdSize(DIAL_WIDTH);
     external.speedDial.setSdCentered(SD_CENTER);
-    setEngine();
     alignPage();
     $('#fadeOverlay2').fadeOut('slow');
     location.reload();
@@ -1007,10 +995,10 @@ function disableCentering() {
 <div class="SearchBar" id="SearchBar">
     <div class="SearchBar" id="SearchBarToggle" onClick="searchToggle()">
     </div>
-        <form id="SearchForm" action="" method="GET">
-            <input id="SearchInput" name="">
-            <input class="submit" type="submit" value="">
-        </form>
+        <div id="SearchForm">
+            <input id="SearchInput" name="" value="%SEARCH-HINT%" onfocus="searchFocus()" onfocusout="searchRelease()">
+            <input id="submit" type="button" value="" onClick="Search()">
+        </div>
      <div style="width: 36px"></div>
 </div>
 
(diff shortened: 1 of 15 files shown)


nowrep:
It looks good, but the configuration needs to be changed.

Either it should offer the same engines as already configured in SearchEnginesManager or just use the default search engine to not complicate things. It can't be like it is now - hardcoded engines.

src/lib/network/schemehandlers/qupzillaschemehandler.cpp
```diff
@@ -294,7 +294,7 @@ QString QupZillaSchemeReply::speeddialPage()
         dPage.replace(QLatin1String("%JQUERY-UI%"), QLatin1String("qrc:html/jquery-ui.js"));
         dPage.replace(QLatin1String("%LOADING-IMG%"), QLatin1String("qrc:html/loading.gif"));
         dPage.replace(QLatin1String("%IMG_SETTINGS%"), QLatin1String("qrc:html/configure.svg"));
-
```
nowrep: Don't remove newline here.

src/lib/plugins/speeddial.cpp
```diff
@@ -318,6 +343,13 @@ void SpeedDial::setSdSize(int count)
     m_sizeOfSpeedDials = count;
 }
 
+void SpeedDial::setsearchEnabled(bool searchenabled)
```
nowrep: `setSearchEnabled`

src/lib/data/html/speeddial.html
```diff
@@ -528,20 +608,31 @@
         $('#fadeOverlay').click();
 });
 
-function emitChanged(pages)
-{
+function searchToggle() {
+    if ("%SEARCH-ENABLED%" == '1') {
+        external.speedDial.setsearchEnabled(false);
```
nowrep: This can be just `external.speedDial.setsearchEnabled("%SEARCH-ENABLED%" == '1')`.

Actually, `%SEARCH-ENABLED%` should be bool and not string.

src/lib/network/schemehandlers/qupzillaschemehandler.cpp
```diff
@@ -331,8 +331,10 @@ QString QupZillaSchemeReply::speeddialPage()
     page.replace(QLatin1String("%IMG_BACKGROUND%"), dial->backgroundImage());
     page.replace(QLatin1String("%URL_BACKGROUND%"), dial->backgroundImageUrl());
     page.replace(QLatin1String("%B_SIZE%"), dial->backgroundImageSize());
+    page.replace(QLatin1String("%SEARCH-ENGINE%"), dial->searchEngine());
     page.replace(QLatin1String("%ROW-PAGES%"), QString::number(dial->pagesInRow()));
     page.replace(QLatin1String("%SD-SIZE%"), QString::number(dial->sdSize()));
+    page.replace(QLatin1String("%SEARCH-ENABLED%"), QString::number(dial->searchEnabled()));
```
nowrep: `QVariant(dial->searchEnabled()).toString()`

nowrep (changes requested):

src/lib/data/html/speeddial.html
Diablopc:
```
function searchToggle() {
    if ("%SEARCH-ENABLED%" == 'true') {
        external.speedDial.setSearchEnabled(false);
    } else {
        external.speedDial.setSearchEnabled(true);
    };
    location.reload();
}
```

nowrep:
The settings have to be in speed dial. Still, I'm okay with it just using default search engine and go without config.

Sorry for not being clear how the settings should be done, next time it would be great if we can discuss it eg. on IRC to prevent this kind of situation.

src/lib/data/html/speeddial.html
```diff
@@ -528,20 +603,26 @@
         $('#fadeOverlay').click();
 });
 
-function emitChanged(pages)
-{
+function searchToggle() {
+    if ("%SEARCH-ENABLED%" == 'true') {
```
nowrep: This can be just

    external.speedDial.setSearchEnabled(%SEARCH-ENABLED%);

Also, move the search enable variable to top as with all other configurable vars: `var SEARCH_ENABLED = %SEARCH-ENABLED%;`, so this line just becomes:


    external.speedDial.setSearchEnabled(SEARCH_ENABLED);

src/lib/data/html/speeddial.html
```diff
@@ -903,19 +988,68 @@
...
+    document.getElementById('quickdial').style.margin = '10px auto';
+    document.getElementById('SearchBar').style.width = '40px';
+    document.getElementById('SearchBar').style.background = 'rgba(0,0,0, 0.0)';
+    document.getElementById('sett').style.top = '4px';
+    document.getElementById('SearchBarToggle').style.background = 'url("qrc:html/search.svg") no-repeat center center, rgba(65,65,65, 0.0)';
+    document.getElementById('SearchForm').style.visibility = 'hidden';
+}
+
+function searchFocus() {
+    document.getElementById('SearchInput').style.color = 'rgba(236,236,236, 1.0)';
```
nowrep: This should be in stylesheet

src/lib/data/html/speeddial.html
```diff
@@ -903,19 +988,68 @@
...
+    document.getElementById('SearchInput').value = '';
+}
+
+function searchRelease() {
+    document.getElementById('SearchInput').style.color = 'rgba(118,118,118, 1.0)';
+    document.getElementById('SearchInput').value = "%SEARCH-HINT%";
```
nowrep: Move %SEARCH-HINT% to variables same as %SEARCH-ENABLED%.

src/lib/data/html/speeddial.html
```diff
@@ -903,19 +988,68 @@
...
+}
+
+function Search() {
+    if (document.getElementById('SearchInput').value == "%SEARCH-HINT%") {
+        value = ''
+    } else {
+        value = document.getElementById('SearchInput').value
+    }
+    var url = '%SEARCH-URL%'.replace('%s', value);
```
nowrep: You can't expose search url to JavaScript as there is also support for `POST` search engines which will not work this way.
You should just expose `search(string)` C++ function to SpeedDial, which will handle the searching.

src/lib/data/html/speeddial.html
```diff
@@ -903,19 +988,68 @@
...
+    this.location.href = url;
+}
+
+document.getElementById('SearchInput').onkeypress = function(e){
+    if (!e) e = window.event;
```
nowrep: Can this happen?

src/lib/data/html/speeddial.html
```diff
@@ -903,19 +988,68 @@
...
+    if (e.keyCode == '13')
```
nowrep: Please add comment that 13 is enter.

nowrep (changes requested):

src/lib/data/html/speeddial.html
Diablopc: It can't be `setSearchEnabled(SEARCH_ENABLED);`
but it can be `setSearchEnabled(!SEARCH_ENABLED);`, this returns value reverse to existing

Changed to:
```
function searchToggle() {
    external.speedDial.setSearchEnabled(!SEARCH_ENABLED);
    location.reload();
}
```

src/lib/data/html/speeddial.html
Diablopc: Yep, but second value need to be changed here, and i see no reason to split them to different code blocks.
There is no problem to move this to stylesheet, but why it needed?

nowrep: Because style should be separate from logic.
Also, as those functions will then be just one statement, you can delete them and move the code inline in html tag.

src/lib/data/html/speeddial.html
nowrep: Sure

src/lib/data/html/speeddial.html
Diablopc: > POST search engines 

I completely forgot about them %)
> which will not work this way.

Why not? It can be simply converted to GET request...
```
.........
    page.replace(QLatin1String("%SEARCH-POST%"), dial->searchEngine().postData);
.........
```
```
.........
var SEARCH_POST = "%SEARCH-POST%";
.........
    if (SEARCH_POST) {
        SEARCH_URL = SEARCH_URL + '?' + SEARCH_POST; 
    }
.........
```
Or as a variant we can generate  different requests depended on engine method.

nowrep: No, this is not how you make POST request from JavaScript. One way to do it is to create temporary `<form>` and accept it.

In any case, that logic is already implemented and tested on C++ side so there is no reason to duplicate it in speed dial.

Diablopc:
> The settings have to be in speed dial. 

I see no reason to return them to SD.
In the current state all search settings are displayed in one place
> Still, I'm okay with it just using default search engine and go without config.

But while use just default search engine we can't make triple search available(
and additional search bar in SD have no reason

> next time it would be great if we can discuss it eg. on IRC to prevent this kind of situation.

Hah, yea. Great idea for future))

nowrep:
All speed dial options are configurable from inside speed dial, so this will not be an exception.

If you really want to use multiple search engines, you have search bar that lets you switch engines quickly with up/down keys. This is really not a reason for adding it to speed dial.

Also this use case has another flaw as if you open new tab with speed dial, you will get locationbar automatically focused so in order to switch to speed dial search bar you either need to either click on the search bad with mouse or (if the stars align) press <kbd>TAB</tbd>.

So if you really want to have it configurable, it needs to be configurable form speed dial page. That means you need to expose available search engines from C++ to Speed Dial and make the search from C++ too (this is needed even if it is not configurable).

nowrep:
If this is still relevant, please submit the patch to Falkon.

PR #2559, opened 2018-01-18
All copyrights at once =)

Diablopc:


nowrep:
I only update year when actually editing the file.

PR #2566, opened 2018-01-24
TabManager: Add Unload checked tabs action

SGOrava:


src/plugins/TabManager/tabmanagerwidget.cpp
```diff
@@ -471,7 +472,7 @@ void TabManagerWidget::processActions()
                 continue;
             }
 
-            if (command == "closeSelection") {
+            if (command == "closeSelection" || command == "unloadSelection") {
```
nowrep: Umm yeah, this is `if (A) { action1() } else if (B) { action1(); }` :D @srazi

SGOrava: It is strange construction, but for this small update i did not want to change a lot of things...

srazi: Sometimes we code and sometimes we do magic :D

I'll fix it, ASAP.

srazi: :D ea0d9289e38bbc57d00c0c082d4baeed1da00466

nowrep: What's wrong with that commit?

srazi: Before that commit actions of `if`s were different.
See this [part](https://github.com/QupZilla/qupzilla/commit/ea0d9289e38bbc57d00c0c082d4baeed1da00466#diff-76d20d37d43595d570512a1e29eeb8bbR474)

nowrep: Yeah right, sorry 😆

PR #2594, opened 2018-02-08
Bookmarks ToolBar refresh fix

TARPTAEYA:


src/lib/bookmarks/bookmarkstoolbar.h
```diff
@@ -57,6 +57,8 @@ private slots:
 
     void dropEvent(QDropEvent* e);
     void dragEnterEvent(QDragEnterEvent* e);
+    
+    int fixedMinHeight = -1;
```
nowrep: Style: `m_` prefix, also it should be added at the end.

src/lib/bookmarks/bookmarkstoolbar.cpp
```diff
@@ -213,8 +213,13 @@ void BookmarksToolbar::addItem(BookmarkItem* item)
     button->setShowOnlyIcon(m_bookmarks->showOnlyIconsInToolbar());
     button->setShowOnlyText(m_bookmarks->showOnlyTextInToolbar());
     m_layout->addWidget(button);
-
-    setMinimumHeight(minimumSizeHint().height());
+    
+    if(fixedMinHeight != -1){
```
nowrep: Style `if (fixedMinHeight != -1) {`

TARPTAEYA:
yeah, only one condition would be sufficient, fixed :)

src/lib/bookmarks/bookmarkstoolbar.cpp
```diff
@@ -213,8 +213,11 @@ void BookmarksToolbar::addItem(BookmarkItem* item)
     button->setShowOnlyIcon(m_bookmarks->showOnlyIconsInToolbar());
     button->setShowOnlyText(m_bookmarks->showOnlyTextInToolbar());
     m_layout->addWidget(button);
-
-    setMinimumHeight(minimumSizeHint().height());
+    
+    if(!m_fixedMinHeight){
```
nowrep: Spaces: `if (!m_fixedMinHeight) {`

TARPTAEYA: Yeah fixed, seems like I must go through code formatting first :grin:

PR #2597, opened 2018-02-09
Fixed ToolTip on Pinning/Unpinning and min height for tabbar

TARPTAEYA:
Here I have tried to fix two bugs
- When only pinned tabs there in a session and when the session is restored then the height of tabbar is too small to fit the tabs, so increased that
- Pinning/ Unpinning tabs failed to show tabToolTip in the current session so fixed that

### DIFF
--- a/src/lib/tabwidget/combotabbar.cpp
+++ b/src/lib/tabwidget/combotabbar.cpp
@@ -534,8 +534,9 @@ void ComboTabBar::setUpLayout()
     }
 
     // We need to setup heights even before m_mainTabBar->height() has correct value
-    // So lets just set minimum 5px height
-    height = qMax(5, height);
+    // So lets just set minimum 32px height so that if just pinned tabs were present
+    // in the last session, then height becomes very small
+    height = qMax(32, height);
 
     setFixedHeight(height);
     m_pinnedTabBar->setFixedHeight(height);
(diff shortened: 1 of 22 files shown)


src/lib/tabwidget/tabstackedwidget.cpp
```diff
@@ -262,6 +262,8 @@ int TabStackedWidget::pinUnPinTab(int index, const QString &title)
 
     bool makePinned = index >= m_tabBar->pinnedTabsCount();
     QWidget* button = m_tabBar->tabButton(index, m_tabBar->iconButtonPosition());
+    // To show tooltip of tab which is pinned in the current session
+    QString m_tabToolTip = tabToolTip(index);
```
nowrep: `m_` prefix is only for member variables, so in this case it should just be `tabToolTip`.

src/lib/tabwidget/combotabbar.cpp
```diff
@@ -534,8 +534,9 @@ void ComboTabBar::setUpLayout()
     }
 
     // We need to setup heights even before m_mainTabBar->height() has correct value
-    // So lets just set minimum 5px height
-    height = qMax(5, height);
+    // So lets just set minimum 32px height so that if just pinned tabs were present
+    // in the last session, then height becomes very small
+    height = qMax(32, height);
```
nowrep: I'm not really convinced by this. Although it does "fix" the issue in most cases, it is still just hardcoded value and will break in case where the real tabbar height is much bigger (so 32 is still too small).

PR #2599, opened 2018-02-11
removed unnecessary cursor:pointer from the body of speeddial

TARPTAEYA:
The current speed dial of Qupzilla shows cursor as pointer even when there is no speeddial box under it. This creates a consufion of a link there sometimes. - So I removed that preserving "cursor:pointer" on other elements


Martii:
Tested between Linux and OS X with eeb69b4 ... Linux shows a finger *(link)* pointer and OS X shows an arrow pointer on the actual thumbnail content of the web page. Don't know if this makes a difference.

PR #2602, opened 2018-02-12
Tabs drag/drop into the same window

TARPTAEYA:
Currently drag/drop in current window just returns, instead It can function like what is done in firefox or chrome that tabs scroll (or placed) within the same window at a  new index.

### DIFF
--- a/src/lib/tabwidget/tabbar.cpp
+++ b/src/lib/tabwidget/tabbar.cpp
@@ -642,7 +642,7 @@ void TabBar::dragMoveEvent(QDragMoveEvent *event)
     const int index = tabAt(event->pos());
     const QMimeData* mime = event->mimeData();
 
-    if (index == -1 || (mime->hasFormat(MIMETYPE) && event->source() == this)) {
+    if (index == -1) {
         ComboTabBar::dragMoveEvent(event);
         return;
     }
@@ -679,10 +679,6 @@ void TabBar::dropEvent(QDropEvent* event)
     }
 
     event->acceptProposedAction();
-
-    if (mime->hasFormat(MIMETYPE) && event->source() == this) {
-        return;
-    }
 
     TabBar *sourceTabBar = qobject_cast<TabBar*>(event->source());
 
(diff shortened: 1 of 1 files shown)


nowrep:
This will detach -> attach the tab instead of just moving it inside the tabbar which will result in changing of current tab and will eg. load unloaded tab.
It should use the same logic as TabModel.

TARPTAEYA:
hmm., I will improve this

TARPTAEYA:
What about this ... ?

nowrep:
Still wrong:
- uses incorrect target index (see prepend/apend tab action)
- doesn't work at all for pinned tabs

If you really want to implement it, it should use the same logic as moving tabs in TabModel (see `TabModel::dropMimeData`).
Although, that being said, moving pinned tabs requires special handling, so in order to not duplicate code, maybe `TabWidget::moveTab` should handle the pinned tab case instead of manually doing it in both TabModel and TabBar.

Pinned tab case: In order to move pinned tab to normal tabs / move normal tab to pinned tabs you first need to pin/unpin them and only after this move them to target index.

TARPTAEYA:
I think this will work now.

nowrep:
Have you seen the logic in TabModel?

```cpp
        if (tab->isPinned()) {
            if (row > m_window->tabWidget()->pinnedTabsCount()) {
                tab->togglePinned();
            }
        } else {
            if (row < m_window->tabWidget()->pinnedTabsCount()) {
                tab->togglePinned();
                row++;
            }
        }
        tab->moveTab(row > mimeData->tab()->tabIndex() ? row - 1 : row);
```

It's this simple and easy to understand, you don't need anything more complex.

Also this code is in wrong place and thus duplicating a lot of code that is done down in the function. Case of moving tab should be handled in:

```cpp
} else if (mime->hasFormat(MIMETYPE) && sourceTabBar) {
    // here
    WebTab *tab = sourceTabBar->webTab();
    if (tab) {
        sourceTabBar->m_tabWidget->detachTab(tab);
```

nowrep:
See e0cbec5a5f6eb5de8edde631528dcc3848048622

PR #2604, opened 2018-02-12
fixed drop url over unrestored tabs

TARPTAEYA:
If the tab is not restored then, restore it and load the dropped url

### DIFF
--- a/src/lib/tabwidget/tabbar.cpp
+++ b/src/lib/tabwidget/tabbar.cpp
@@ -708,8 +708,9 @@ void TabBar::dropEvent(QDropEvent* event)
             req = mApp->searchEnginesManager()->searchResult(mime->text());
         }
         if (action == SelectTab) {
-            if (tab->isRestored() && req.isValid()) {
-                tab->load(req);
+            if (req.isValid()) {
+                tab->tabActivated();
+                QTimer::singleShot(300, this,[tab, req]{tab->load(req);});
             }
         } else if (action == PrependTab || action == AppendTab) {
             const int newIndex = action == PrependTab ? index : index + 1;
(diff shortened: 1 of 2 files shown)


nowrep:
Oh and most important: This should be handled in `WebTab::load`, not here.

src/lib/tabwidget/tabbar.cpp
```diff
@@ -708,8 +708,9 @@ void TabBar::dropEvent(QDropEvent* event)
             req = mApp->searchEnginesManager()->searchResult(mime->text());
         }
         if (action == SelectTab) {
-            if (tab->isRestored() && req.isValid()) {
-                tab->load(req);
+            if (req.isValid()) {
+                tab->tabActivated();
+                QTimer::singleShot(300, this,[tab, req]{tab->load(req);});
```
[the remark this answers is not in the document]
TARPTAEYA: 300 is just a timeout value, because if this occurs immediately then the tab gets loaded before restoring, and thus when index changed it restores

TARPTAEYA:
Done

src/lib/webtab/webtab.cpp
```diff
@@ -385,6 +385,10 @@ void WebTab::reload()
 
 void WebTab::load(const LoadRequest &request)
 {
+    if (!isRestored()) {
+        tabActivated();
+        QTimer::singleShot(300, this, std::bind(&WebTab::load, this, request));
+    }
```
nowrep: `else`

src/lib/webtab/webtab.cpp
```diff
@@ -385,6 +385,10 @@ void WebTab::reload()
 
 void WebTab::load(const LoadRequest &request)
 {
+    if (!isRestored()) {
+        tabActivated();
+        QTimer::singleShot(300, this, std::bind(&WebTab::load, this, request));
```
nowrep: Can you try only with QTimer::singleShot(0)?

Although maybe clean solution is to just remove the singleShot(0) call from tabActivated, I'm not sure if it really is needed at all there.

TARPTAEYA: yeah... it worked with singleShot(0)

PR #2605, opened 2018-02-13
prevent duplicate bookmarks

TARPTAEYA:
context menu and drop url creates duplicate bookmarks if done more than once because there is no check if the bookmark already exists.

### DIFF
--- a/src/lib/bookmarks/bookmarksmodel.cpp
+++ b/src/lib/bookmarks/bookmarksmodel.cpp
@@ -51,6 +51,10 @@ void BookmarksModel::addBookmark(BookmarkItem* parent, int row, BookmarkItem* it
     Q_ASSERT(row >= 0);
     Q_ASSERT(row <= parent->children().count());
 
+    if (m_bookmarks->isBookmarked(item->url())) {
+        return;
+    }
+
     beginInsertRows(index(parent), row, row);
     parent->addChild(item, row);
     endInsertRows();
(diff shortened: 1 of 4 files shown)


nowrep:
I don't see a reason why to disallow multiple bookmarks with same url.
Also the code calling this method (`BookmarksModel::addBookmark`) assumes that adding bookmark always succeeds, so this will break this assumption.

TARPTAEYA:
It fills the space of bookmark toolbar with same bookmark otherwise

nowrep:
There already is a warning that the url is bookmarked when creating new bookmark.

Martii:
> There already is a warning that the url is bookmarked when creating new bookmark.

There is? Just tried this with the address bar, and a context click with no visible warnings to my knowledge:

<a href="https://user-images.githubusercontent.com/114709/36234240-261d45d4-11a7-11e8-9d84-10bde275b217.png">bookmarking end result</a>

nowrep:
Hmm I thought there is, but there really is not.

In that case a warning should be added to add bookmark dialog (`BookmarksTools::addBookmarkDialog`).

nowrep:
This now lets the user modify the bookmark title / folder and only show warning messagebox right before actually storing the bookmark. That's just annoying and late (+ it leaks BookmarkItem in case of bookmark not being added).

What I had in mind with "warning" is to just show text in add bookmark dialog like (wording can be changed):

**NOTE:** This site is already bookmarked.

This label can be added eg. between folder combobox and dialog button box (ok/cancel buttons).

TARPTAEYA:
Done, but I have preserved the `BookmarksTools::allowDuplicateBookmarks` for the case when dropping duplicate urls to bookmarks toolbar

Issue #2608, opened 2018-02-14
Feature  request: Printing to PDF with the page title


test666v2:
**Feature  request: Printing to PDF with the page title**

Printing to PDF now defaults to "print.pdf". If possible, please restore the capabilitiy to automatically insert the page title as the filename.

Keep up the good work. Thank you.

TARPTAEYA:
created a PR for the same.

nowrep:
5372944860da8b93d9cd755f3cb782a4122b6181

PR #2609, opened 2018-02-15
resloved issue 2608 - Printing to PDF with the page title

TARPTAEYA:


Martii:
Ref backlink to #2608

nowrep:
This is still using url and not page title as requested in #2608

TARPTAEYA:
fixed

nowrep:
What name it uses when title is `foo/bar/123\456`?

TARPTAEYA:
done

src/lib/webengine/webview.cpp
```diff
@@ -392,14 +392,16 @@ void WebView::printPage()
 {
     Q_ASSERT(m_page);
 
-    QPrintDialog* dialog = new QPrintDialog(this);
+    QPrinter* printer = new QPrinter(QPrinter::HighResolution);
```
nowrep: Any reason for `QPrinter::HighResolution` instead of default?

TARPTAEYA: No specific reason, reverted it back to default.

PR #2621, opened 2018-02-25
clear mouseover state after closing menu in bookmarks toolbar

TARPTAEYA:


nowrep:
This is rebuilding the entire toolbar every time the context menu is closed.
`QMenu::popup()` may be enough to fix the issue.

